PR #76, opened 2024-08-23
docs: Update README.md to fix the build & test badge

PeterMarinov:
Fixing the broken Readme after renaming the build & test workflow


orilux (approved):
The updated url works for me.

PR #88, opened 2024-08-30
feat: Adding Next.js app following the agreed directory structure

PeterMarinov:
# Adding Next.js app following the agreed directory structure

## Description
We will use `apps/endatix-hub` to setup the Endatix management UI.

To run the app:
1. cd into `apps/endatix-app`
2. run `npm install`
3. run `npm run dev`

## Related Issues
unblocks #84 

## Type of Change


apps/endatix-hub/services/sitemap-service.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!excludeSettings){
+      sitemapList.push(sitemap.Settings);
+    }
+
+    return sitemapList;
+  }
+
+  public static getLogo(): INavItem {
+    const logoNavItem: INavItem = {
+      path: "https://surveysocket.com",
```
dspetrov, line 35: This and the below are left with the old name.

dspetrov (changes requested):
LGTM, I've added a comment considering an old name left.

apps/endatix-hub/services/sitemap-service.ts
PeterMarinov: Great for spotting this. Let me fix in the next commit

PeterMarinov:
@dspetrov this is done. Also added the auth WIP piece. Step 1 to 3 are done, but stopping here to keep scope clean:
image

dspetrov (approved):

PR #96, opened 2024-09-13
docs: Fix URLs in docker/README.md

dspetrov:
# Fix URLs in docker/README.md

## Description
There are a few URLs in docker/README.md that are wrong and need to be fixed.


PeterMarinov (approved):
@dspetrov  - Urls are working. Merging this one.

PR #104, opened 2024-09-18
fix: Change DB and API healthcheck config, remove healthz requests from logs

### LINKED ISSUE #100: [BUG] Endatix API log becomes too large
**Describe the bug**
Endatix API log becomes too large too quickly because of too many info messages connected with healthcheks to /healthz endpoint.

**To Reproduce**
Leave Endatix API running in a Docker container for a few hours and there will be 10s of thousands of log messages for /healthz.

**Expected behavior**
The log contains helpful information for the application's execution and issues it has.

**Desktop:**
 - Running Endatix API in a Docker container

**Proposed solution**
- the Endatix API Docker container health check interval is too often. Change the endatix-api healthcheck in docker-compose files to
      interval: 15s
      retries: 3
      start_period: 10s
      timeout: 3s
- although not connected with the API log, the MS SQL Server Docker container health check interval is also too often. Change the endatix-db healthcheck in docker-compose files to
      interval: 15s
      retries: 3
      start_period: 20s
      timeout: 3s
- There is no need to log as info all the calls to healthz. This can possibly be achieved by changes in the level when defining `UseSerilogRequestLogging`, e.g.:
```
app.UseSerilogRequestLogging(options =>
{
    options.GetLevel = (httpContext, elapsed, ex) =>
        httpContext.Request.Path.StartsWithSegments("/healthz")
            ? LogEventLevel.Verbose // Log successful health checks with low verbosity
            : LogEventLevel.Information; // Default level for other requests
});
```
This is a proposal and needs a confirmation how it works. It is important to skip ONLY the /healthz logging and ONLY the success calls. In case of errors during /healthz calls, the messages MUST be logged.

dspetrov:
# Change DB and API healthcheck config and remove /healthz requests from logs

## Description
Logs become large very fast becaus of logging /healthz requests every 5 s. To prevent this and to make the operation of Endatix API smoother, two things are done:
- Extend the healthcheck interval to15 sec (including a few more healthcheck config values changed that are better now)
- Changes in the Serilog setup to skip /healthz requests from logging and also change the logging level of Microsoft/System logging on dev so it doesn't automatically log all the requests (this is already done by Serilog)

## Related Issues
closes #100

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
+++ b/src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
@@ -26,9 +26,14 @@ public static class AppBuilderExtensions
         app.UseSerilogRequestLogging(options =>
         {
             options.GetLevel = (httpContext, elapsed, ex) =>
-                httpContext.Request.Path.StartsWithSegments("/healthz") ?
-                    LogEventLevel.Verbose :
-                    LogEventLevel.Information;
+            {
+                if (httpContext.Request.Path.StartsWithSegments("/healthz"))
+                {
+                    return LogEventLevel.Verbose;
+                }
+
+                return LogEventLevel.Information;
+            };
         });
 
         app.UseHttpsRedirection();
--- a/src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
+++ b/src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
@@ -2,6 +2,7 @@ using Endatix.Framework.Hosting;
 using Endatix.Setup;
 using Microsoft.AspNetCore.Builder;
 using Serilog;
+using Serilog.Events;
 
 namespace Endatix.Setup;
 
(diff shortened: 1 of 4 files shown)


src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
```diff
@@ -22,7 +23,13 @@ public static IEndatixMiddleware UseEndatixMiddleware(this WebApplication app)
         app.UseAuthentication()
             .UseAuthorization();
 
-        app.UseSerilogRequestLogging();
+        app.UseSerilogRequestLogging(options =>
```
PeterMarinov, line 26: It's not great that we cannot use appSettings for this change, but if that's what it takes I am OK.

One code style suggestion: imo this code is a big hard on the eye. How about separate statements for readability?

```c#
options.GetLevel = (httpContext, elapsed, ex) =>
{
    if (httpContext.Request.Path.StartsWithSegments("/healthz"))
    {
        return LogEventLevel.Verbose;
    }

    return LogEventLevel.Information;
};
```

PeterMarinov (approved):
Suggested one chage

src/Endatix.Extensions.Hosting/AppBuilderExtensions.cs
dspetrov: It shouldn't be a problem that it's not configurable through appSettings as this logging is not needed. If due to some reason this is necessary, it can always but turned on from the settings by setting the level temporarily to Verbose.
It doesn't seem hard to read for me as it is a simple use of the ternary conditional operator but I'll change it to if block so it is clearer.

PeterMarinov (approved):

Issue #33, opened 2024-07-29
Add Clean Validation pattern


dspetrov:
**Original author:** @PeterMarinov 
**Original URL:** https://trello.com/c/77kQID4t/34-add-clean-validation-pattern

To avoid adding repetitive validation code all across the application, we need to clean up how it’s being used and set a pattern. Possible solution will be the FluentValidation + validation pipeline via MediatR, so it’s repeatable with no additional code needed

### Acceptance Criteria
- Implement Fluent Validation
- Add ValidationBehavior to the Mediator behavior pipeline
- Implement validation for the Commands already implemented
- Add tests on the validation
- Add documentation article

### Resources
- https://www.youtube.com/watch?v=9KuLsPV8BYU#sq_hpzspg0fvx
- https://github.com/PeleaCezar/Gatherly/blob/b6366a80a74a343007317435f2bd3e1227794212/Gatherly.Application/Behaviors/ValidationPipelineBehavior.cs - by Milko Javanovich
- https://www.youtube.com/watch?v=dpPcnAT7n7M

PR #125, opened 2024-09-27
feat: Change identity ID type to long #107

### LINKED ISSUE #107: Change identity ID type to `long`
>[!IMPORTANT]
>This must be implemented with using the `feature/46-integrate-identity` as a feature branch. You will branch from it and when ready simply make a PR back to the `feature/46-integrate-identity`

### Goal
The ID of the `AppUser` persisted in the DB is of type string. We must convert it to `long` to match the Ids of the other entities

### Scope
* Change the type of the AppUser's Id prop to `long`
* Generate new from scratch Db migrations
* Migrate old users to the new Id type strategy `out-of-scope`
* Partial migration of current DbMigration `out-of-scope` - since we don't have the initial migration pushed to `main` yet

### Acceptance Criteria
* New users registered will have their Id being of Long type
* Long is generated via the `IIdGenerator` interface
* Authentication works for newly registered users using valid credentials 

### Deliverables
* Working functional code
* New Migration

### Proposed Solution
* Follow the official guide - https://learn.microsoft.com/en-us/aspnet/core/security/authentication/customize-identity-model?view=aspnetcore-8.0#change-the-primary-key-type
* To run migrations use this command `dotnet ef migrations add InitialIdentity --startup-project src/Endatix.WebHost --project src/Endatix.Persistence.SqlServer  --context AppIdentityDbContext --output-dir Migrations`

PeterMarinov:
# Change identity ID type to long 

## Description
- Change the type of the `AppUser.Id` prop to long
- Added the same for `IdentityRoles` and added `AppRole` class to wrap it and allow us to extend `bonus`
- Both Ids are now generated via the `IIdGenerator` interface
- Generated new from scratch Db migrations
- Extended the `IIdGenerator` interface to require ID type, so it's more compliant with EF Core and the upcoming UUID7 of type GUID. Now we have `IIdGEnerator<TKey>` and implementation is coupled with `IIdGEnerator<long>`, but much easier to extend in the future when the time comes
- Refactored AppDbContext to extract that should be reusable. Now it's in `src/Endatix.Framework/Platform/AssemblyExtensions.cs` 

## Related Issues
addresses #107 

## Type of Change

## Checklist

## 💪  Setup Instructions

> [!IMPORTANT]
> Run this code below to drop the identity tables and the DB migrations, so you can start fresh. This needed as the feature is not on `main` and will be best not waste dev time to implement better migration script

```sql
IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[__EFMigrationsHistory]') AND type in (N'U'))
DROP TABLE [dbo].[__EFMigrationsHistory]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUserRoles]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUserRoles]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUserTokens]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUserTokens]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUserLogins]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUserLogins]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUserClaims]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUserClaims]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUserClaims]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUserClaims]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetRoleClaims]') AND type in (N'U'))
DROP TABLE [identity].[AspNetRoleClaims]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetRoles]') AND type in (N'U'))
DROP TABLE [identity].[AspNetRoles]
GO

IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[identity].[AspNetUsers]') AND type in (N'U'))
DROP TABLE [identity].[AspNetUsers]
GO
```

## Screenshots
image


src/Endatix.Core/Entities/Identity/Role.cs
```diff
@@ -0,0 +1,36 @@
...
+/// </summary>
+public sealed class Role : BaseEntity, IAggregateRoot
+{
+    public Role(
+        long id,
+        string name,
+        string? description
+        )
+    {
+        Guard.Against.Null(id);
```
dspetrov, line 20: This check doesn't seem to work because id is not nullable.
In addition, I think we expect having only positive values for IDs and in other places we have guard clauses using `NegativeOrZero` for IDs. I suggest we use it here too.
The same applies for ID in User class.

Another question that comes to my mind is if we guard the ID, shouldn't we guard also name with NullOrWhiteSpace as having such a role/user is not correct.

src/Endatix.Framework/Platform/AssemblyExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    /// <summary>
+    /// Get all Endatix platform assemblies
+    /// </summary>
+    /// <param name="assembly">The assembly to start from</param>
+    /// <returns>A HashSet of Endatix platform assemblies to ensure no duplication of assemblies</returns>
+    public static HashSet<Assembly> GetEndatixPlatformAssemblies(this Assembly assembly)
+    {
+        HashSet<Assembly> endatixAssemblies = [];
+        if (assembly == null || !assembly.IsEndatixAssembly())
```
dspetrov, line 18: I don't understand the purpose of this check. Firstly, do we expect `GetEndatixPlatformAssemblies` to be called on `null`, as it is an extension method?
The second part is also not quite clear to me - we don't wish to continue if the current assembly is not an Endatix assembly even if it's another product that references Endatix?

dspetrov (changes requested):
Looks good and I just have a couple of small things to clarify and possibly change. Please check the comments.

src/Endatix.Core/Entities/Identity/Role.cs
PeterMarinov: @dspetrov  - you are right. We should check for what's critical. I have added additional Guard clauses and fixed the Id one. ✅ resolved

Side note - Need to check at some point for more integrated validation for validating commands passed between MediatR. I have checked this - since we use Result, we can output Result.Invalid() with the validation errors before we process the items in the pipeline. Here is the task- https://github.com/endatix/endatix/issues/33

dspetrov (approved):

Issue #114, opened 2024-09-20
Add support for refresh token


PeterMarinov:
>[!IMPORTANT]
>This must be implemented with using the `feature/46-integrate-identity` as a feature branch. You will branch from it and when ready simply make a PR back to the `feature/46-integrate-identity`

### Goal
We want to allow clients of the API to be able to use Refresh tokens to renew their JWT tokens without the need to send their credentials again

### Scope
* Add refresh token endpoint
* Add revoke token endpoint `out-of-scope`
* Modify authorization logic to return 401 with token_expired error as (if) required - (according to the spec rfc6750 - "The OAuth 2.0 Authorization Framework: Bearer Token Usage", https://www.rfc-editor.org/rfc/rfc6750, p.8, section 3.1, resource server should return 401)
* Update config values as required e.g. add refresh token lifetime value

### Acceptance Criteria
* Users with valid, but expired JWT token can make a call to the refresh endpoint and receive a new token
* Anonymous users cannot access the refresh token endpoint
* Refresh tokens renew JWT token only of the JWT token is valid and the Refresh token is valid and they belong to the same user
* Refresh tokens can only be used once
* Users with expired tokens receive 401 error with token expired error
* Revoke token endpoint invalidates the refresh token
* Refresh request with revokend refresh token returns an error
* Users can have no more than one refresh token at a time

### Deliverables
* Working endpoints
* Tests

### Proposed Solution
* Check the https://github.com/dotnet/aspnetcore/blob/main/src/Identity/Core/src/IdentityApiEndpointRouteBuilderExtensions.cs#L122 and review the implementation. We don't use the SignInManager but have access to the UserStore and UserManager, so we should be able to have the same logic
* Latest version FastEndpoints has a new feature to support token revocation and refresh token - https://fast-endpoints.com/docs/security#jwt-token-revocation. Might be helpful - https://github.com/FastEndpoints/Refresh-Tokens-Demo

PeterMarinov:
@dspetrov  - this is an example of what we chatted about, where the Refresh token is handled as persisted attribute of the `IdentityUser` (`AppUser`) in our case.

Let's continue the feature with having the Refresh token specific logic for persistence, generation, CRUD in the `ITokenService` and maybe use the `IAuthService` for Token validation and auth orchestration logic

PeterMarinov:
@dspetrov , @orilux  - closing this task

PR #127, opened 2024-09-30
feat: Add Logout functionality

### LINKED ISSUE #106: Add signout endpoint

PeterMarinov:
# Add Logout functionality

## Description

- feat: Add logout functionality
- feat: Add additional ToApiEndpoint extension method for simpler flow with no mapper
- feat: Adding default claim for identifying user Id, so that ASP.NET Core Identity can operate as expected
- feat: Adding logout method in the api.http REST client examples

## Related Issues
closes #106 

## Type of Change

## Checklist

## Note

>[!NOTE]
> I haven't added unit tests for this feature, but will make sure to catch up when there's more time. I have added a slot for the token revocation in the `JwtTokenService` for the #114 feature
```c#
/// <inheritdoc />
    public Task<Result> RevokeTokensAsync(User forUser, CancellationToken cancellationToken = default) {
        if (forUser == null){
            return Task.FromResult(Result.NotFound());
        }

        // Add removal of Refresh token once feature is implemented.
        // Add Revoking JWT token should we decide to implement this
        return Task.FromResult(Result.Success());
    }
```

### DIFF
--- a/src/Endatix.Api/Endpoints/Auth/Logout.cs
+++ b/src/Endatix.Api/Endpoints/Auth/Logout.cs
@@ -1,0 +1,45 @@
+using Endatix.Api.Infrastructure;
+using Endatix.Core.Infrastructure.Result;
+using Endatix.Core.UseCases.Identity.Login;
+using Endatix.Infrastructure.Identity.Authorization;
+using FastEndpoints;
+using MediatR;
+using Microsoft.AspNetCore.Http.HttpResults;
+
+namespace Endatix.Api.Endpoints.Auth;
+
+public class Logout(IMediator mediator) : EndpointWithoutRequest<Results<Ok<LogoutResponse>, BadRequest>>
+{
+    /// <summary>
+    /// Configures the endpoint settings for the logout functionality.
+    /// </summary>
+    public override void Configure()
+    {
+        Post("auth/logout");
+        Permissions(Allow.AllowAll);
+        Summary(s =>
+        {
+            s.Summary = "Logs out the authenticated user";
+            s.Description = "Initiates the logout process for the authenticated user.";
+            s.Responses[200] = "User logged out successfully.";
+            s.Responses[400] = "Invalid request or authentication state.";
+        });
+    }
+
+    /// <summary>
+    /// Executes the logout functionality
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token for the async operation.</param>
+    public override async Task<Results<Ok<LogoutResponse>, BadRequest>> ExecuteAsync(CancellationToken cancellationToken)
+    {
+
+        var logoutUserCmd = new LogoutCommand(User);
+        var logoutOperationResult = await mediator.Send(logoutUserCmd, cancellationToken);
+
+        var logoutResponse = logoutOperationResult.IsSuccess ?
+            Result.Success(new LogoutResponse("User logged out successfully.")) :
+            Result.Invalid(new ValidationError("Invalid request or authentication state."));
+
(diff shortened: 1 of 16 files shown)


PeterMarinov:
@dspetrov , @orilux  or whoever is reviewing this PR. Since we only support `JwtBearerToken` scheme we don't have actual logout as the JWT tokens are stateless and not persisted in the DB. They are passed to the API clients and they have to delete them on logout as guide.

### Suggested Logout Strategy

- When we support REfresh tokens, they will be stored in DB, so we will delete them on logout. This will prevent a refresh
- Shorten the JWT duration to 15 minutes. This is recommended practice and we will employ refresh tokens for extending the session of users behind the scenes. This way logged out users will have their access revoked in shorter amount of time
- Endatix Hub already deletes the JWT token on logout from the Hub and we don't share it with the browser, so it's pretty efficient and secure there 🔒 . Just showing off 😎 
- Should we decide to implement token revocation, we will create a DB stored list of "not_allowed_JWT_tokens" and we can check for JWTs in the auth layer. Here's an example - https://fast-endpoints.com/docs/security#jwt-token-revocation.
     - 🟢 Pros is that we have immediate access revocation
     - 🔴 Cons is the overhead on checking the JWT against the DB on every request, so even if we implement this should be optional feature

### Further Read
- [One does not simply log out when using JWT](https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6)
- [Enhancing Website Security: Blocking JWT Token Usage After Logout](https://medium.com/@govindkedia71086/enhancing-website-security-blocking-jwt-token-usage-after-logout-6c54188c95ed)

src/Endatix.Api/Endpoints/Auth/Logout.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Executes the logout functionality
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token for the async operation.</param>
+    public override async Task<Results<Ok<LogoutResponse>, BadRequest>> ExecuteAsync(CancellationToken cancellationToken)
+    {
+
+        var logoutUserCmd = new LogoutCommand(User);
+        var logoutOperationResult = await mediator.Send(logoutUserCmd, cancellationToken);
+
+        var logoutResponse = logoutOperationResult.IsSuccess ?
```
dspetrov, line 39: I know that we cannot directly use the logic of `ToEndpointResponse` as in the majority of our endpoints (e.g. for forms, definitions and submissions) but still I think we can move the logic with checking the status to the command handler.
This way we'll have the logic here cleaner on one side. Something more - currently `Result.Success` receives an object defined in the API assembly and `Result.Invalid` receives and object defined in the Core assembly. It will be good if we are more consistent here. Considering `LogoutResponse` has just a message, we can have a change in the core result to include an informative message for the for the happy path and construct the result object in the command handler with either the result with the message or the validation error.
This is just one proposal and although I haven't tried it, I think it shouldn't be hard. Additionally we can use the approach for other commands that don't return entities but need a massage.
What do you think?

dspetrov (changes requested):
All looks good to me except something I explain in a comment.

src/Endatix.Api/Endpoints/Auth/Logout.cs
PeterMarinov: Yeah, I mutate the validation error by ignoring the original one, which is not very logical and now that I think about it , it's very shady business if done this way. Let me refactor to :
1. Propagate validation error from the original command
2. Encapsulate the message/no message flow within the extension method as you have suggested

dspetrov: Cool, looks much better now! Thanks

dspetrov (approved):

PR #135, opened 2024-10-03
feat: [WIP] Refresh token support

### LINKED ISSUE #114: Add support for refresh token

dspetrov:
# Refresh token support

## Description
A WIP PR for the refresh-token endpoint
Generally it works but the User ID is passed in the refresh-token endpoint payload instead of retrieved from the JWT token. The desired approach needs to have a custom check of the JWT access token in the refresh-endpoint handler - even for an expired JWT token to get the User ID from it if it's a valid one.

The values 15 mins and 7 days are set as default based on research for good standard practices.

## Things that need to be done:
- Implement the above logic
- Check if the correct status codes are returned on expired tokens
- Go though all the PR code and handle errors properly, where not done
- Go though all the PR code and add proper validation, where not done
- Go though all the PR code and add comments, where missing
- Write tests
- Something that looks strange for me is the current implementation of the validity check in Endatix.Core.UseCases.Identity.Login.LoginHandler - it fails on validationResult.IsInvalid() and proceeds otherwise. How about any other not successful statuses? Especially for such essential security endpoint, I suggest we proceed only when the validation result is explicitly checked for success. Same thing is in the new RefreshToken command but it is just copied
- Due to the similarities in the logic of login and refresh-token endpoints, there are several places with code repetition. This should be handled so we follow DRY.

## Related Issues
closes #114

## Type of Change

- [ ] Documentation update

## Checklist
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes


src/Endatix.Infrastructure/Identity/Authentication/JwtTokenService.cs
```diff
@@ -70,6 +72,48 @@ public TokenDto IssueToken(User forUser, string? forAudience = null)
...
+            ValidateLifetime = validateLifetime,
+            ValidateIssuerSigningKey = true,
+            ClockSkew = TimeSpan.FromSeconds(JWT_CLOCK_SKEW_IN_SECONDS)
+        };
+
+        JwtSecurityToken validatedJwtToken;
+        try
+        {
+            var tokenHandler = new JwtSecurityTokenHandler();
+            tokenHandler.ValidateToken(accessToken, validationParameters, out var validatedToken);
```
PeterMarinov, line 93: Not sure if you have checked this, but the `ValidateTokenAsync` will be easier to translate to our business logic as it throws no exceptions at all - https://learn.microsoft.com/en-us/dotnet/api/system.identitymodel.tokens.jwt.jwtsecuritytokenhandler.validatetokenasync?view=msal-web-dotnet-latest.

> Validates a token. On a validation failure, no exception will be thrown; instead, the exception will be set in the returned TokenValidationResult.Exception property. Callers should always check the TokenValidationResult.IsValid property to verify the validity of the result.

tests/Endatix.Api.Tests/Endpoints/Auth/RefreshTokenTests.cs
```diff
@@ -0,0 +1,67 @@
...
+    private readonly RefreshToken _endpoint;
+
+    public RefreshTokenTests()
+    {
+        _mediator = Substitute.For<IMediator>();
+        _endpoint = Factory.Create<RefreshToken>(_mediator);
+    }
+
+    [Fact]
+    public async Task HandleAsync_ValidTokens_ReturnsOkWithNewTokens()
```
PeterMarinov, line 23: Do you think we need a test where the Bearer token is invalid and doesn't start with Bearer to see the expected response? Just a question - not a todo.

tests/Endatix.Infrastructure.Tests/Identity/Authentication/JwtTokenServiceTests.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        // Assert
+        result.Should().NotBeNull();
+        result.Token.Should().NotBeNullOrEmpty();
+        result.ExpireAt.Should().BeAfter(DateTime.UtcNow.AddDays(_validJwtOptions.RefreshExpiryInDays - 1));
+        result.ExpireAt.Should().BeBefore(DateTime.UtcNow.AddDays(_validJwtOptions.RefreshExpiryInDays + 1));
+    }
+
+    [Fact]
+    public async Task RevokeTokensAsync_NullUser_ReturnsNotFoundResult()
```
PeterMarinov, line 221: @dspetrov  - do we really need tests for RevokeTokens? Maybe I am missing something, but the method is a placeholder as of now, so I think these tests will give just false negatives that there a not issues. Maybe we don't need these tests or leave them? In any case, there's no much value in them for now I think. 

Maybe a better approach will be to add implementation for the `RevokeTokensAsync` method and delete the refresh token of the user. What's you take here?

src/Endatix.Api/Endpoints/Auth/Login.cs
```diff
@@ -44,7 +44,7 @@ public override async Task HandleAsync(LoginRequest request, CancellationToken c
         }
         else
         {
-            var successfulResponse = new LoginResponse(request.Email, result.Value.Token, string.Empty);
+            var successfulResponse = new LoginResponse(request.Email, result.Value.AccessToken.Token, result.Value.RefreshToken.Token);
```
PeterMarinov, line 47: I am getting a bit nervious when there are 4 items chained even if payload is validated. Might be good to discuss this and set some guideline . I read more about it and it seems what we can do is to leverage the Nullable reference types more and do something like:
>  `result.Value!.AccessToken!.Token!` - need to dig deeper. Here is a read - https://wildermuth.com/2023/02/13/nullable-reference-types-in-csharp/

Not a TODO again. Just sharing

PeterMarinov (changes requested):
@dspetrov  - review is done. I think it's good to implement revoking the RefreshToken on logout and consider adding Value object for the refresh token before we merge the migrations. Let me know

src/Endatix.Core/UseCases/Identity/RefreshToken/RefreshTokenHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+/// <summary>
+/// Handles the refresh token logic by validating the refresh token and issuing new access and refresh tokens.
+/// </summary>
+public class RefreshTokenHandler(IAuthService authService, ITokenService tokenService) : ICommandHandler<RefreshTokenCommand, Result<AuthTokensDto>>
+{
+    public async Task<Result<AuthTokensDto>> Handle(RefreshTokenCommand request, CancellationToken cancellationToken)
+    {
+        var accessTokenValidationResult = tokenService.ValidateAccessToken(request.AccessToken, validateLifetime: false);
+
+        if (accessTokenValidationResult.IsInvalid())
```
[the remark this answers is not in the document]
dspetrov: Yes, it will be good to have some more general approach that fits the logic in the commands/queries

src/Endatix.Infrastructure/Identity/Authentication/JwtTokenService.cs
dspetrov: Good point, I didn't know with the async method we can get away from handling of an exception, which I definitely didn't like. Thanks.

tests/Endatix.Api.Tests/Endpoints/Auth/RefreshTokenTests.cs
dspetrov: The validator ensures the header is in the correct format. Generally we can test it but so far we don't test the validators logic and I think it's fine to go without this.

tests/Endatix.Infrastructure.Tests/Identity/Authentication/JwtTokenServiceTests.cs
dspetrov: The test passes and yes, it may not provide much value but it doesn't hurt also. It will be extended if the method has a full implementation.

src/Endatix.Api/Endpoints/Auth/Login.cs
dspetrov: I'm not sure if we really need to dig into this. All the values are not nullble, not just because the command ensures they are set correctly but also they are defined as not nullable. Technically we can use the ! operator but it doesn't bring anything and makes the code a bit less cleaner (though just a bit).
Generally, such kind of issues are pointed very clear by the IDE and IMO it we can focus on details that could be problematic for the correctness of how the product works or any general issues with the implementation, especially when they lead to complications with the code maintainability and extension.

src/Endatix.Infrastructure/Identity/AppUser.cs
```diff
@@ -10,6 +10,16 @@ namespace Endatix.Infrastructure.Identity;
 /// </summary>
 public class AppUser : IdentityUser<long>
```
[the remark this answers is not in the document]
dspetrov: I think we don't need such complications. The implementation now works pretty straightforward and simple.
Of course, it is a matter of opinion and both approaches will work.
However, there are at least a few reasons for keeping it this way:
- It's flat and clearer
- In the AuthService (which is where the validation occurs) we have the clear validation logic and conditions
- Though these are knowledge details, I'm not aware how it will work for persisting the record in the DB. It is quite possible that such changes although looking trivial will lead to some more changes needed for the persistence or the further data utilization
- Using DateTime.UtcNow inside this new class sooner or later will bring a dependency as we'll need to use date/time provider. Of course, this is solvable, but is another complication.

I'm not sure what other points can appear as these just come to my mind at first glance.

My general thinking is that we need to keep the things simple and straightforward. Especially at the current stage of the product, having the mindset of focusing on the problematic areas or changing the implementation when it can really make the code maintainability and extension easier, will help us not only with the implementations but also in reviewing PRs and addressing issues.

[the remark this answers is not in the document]
PeterMarinov: I suggest you review this article and reconsider your arguments against using value objects - https://ardalis.com/refactoring-value-objects/. Even if you are not willing now, at least I want to win you for the idea. We can always to it later, but I disagree calling value objects complications. I am convinced that that their are the opposite - a simplification over the complexity which is already showing (and that's just the start of it) as the domain model grows. 


| item | without value objects | with value objects |
| ---- | ----- | ------ |
| - It's flat and clearer | once we have 10+ primitive items, this will turn into disadvantage. I see our service classes grow too much and validation logic spread to many places. Primitive Obsession is an anti-pattern. This article does a good job explaining it - https://enterprisecraftsmanship.com/posts/functional-c-primitive-obsession/ | we can express out model better and ecapsulate logic related to it. It's only complex as we haven't done it yet, not because it's complicated imo
| AuthService | In the AuthService (which is where the validation occurs) we have the clear validation logic and conditions | In this case it will be moved to the RefreshToken and meaning in the AuthService will be just to check if it's valid and not over communicate the actual implementation details of why is valid. This is pure encapsulation, so I see it as an improvement
| - Though these are knowledge details, I'm not aware how it will work for persisting the record in the DB. It is quite possible that such changes although looking trivial will lead to some more changes needed for the persistence or the further data utilization | Not change if we keep it this way | With Value  Objects the column names will be prefixed with the Value Object, e.g. `refreshToken_Hash`, so yes migration will do it in the future. This is why I wanted us to do it now
| - Using DateTime.UtcNow inside this new class sooner or later will bring a dependency as we'll need to use date/time provider. Of course, this is solvable, but is another complication. | Agree, but TimeProvider is adding complexity in any case | Time provider will be only needed for Creation logic, so I foresee we will inject it in CommandHandlers or Services that create Value Objects as part of response to a user command |

PeterMarinov (approved):

PR #136, opened 2024-10-14
feat: Improve support for Results Union Type and Result<T> in Fast Endpoints 

### LINKED ISSUE #31: DevX: Improve support for Specification.Result in FastEndpoints
**Original author:** @PeterMarinov 
**Original URL:** https://trello.com/c/zPB0zHVL/54-devx-improve-support-for-specificationresult-in-fastendpoints

By introducing the `ToEndpointResponse` extension method with https://bitbucket.org/orilux/surveysocket/pull-requests/19, we have been able to greatly improve `Result` handling logic + mapping logic, but there’s a lot more to do here

### Goals
We need to iterate over the way FastEndoints are used to achieve the following goals:
- Improve readability of the code
- Improve dev productivity for faster development of endpoints + testing
- Performance
- Compile time safety

### Current Problems
- Readability is not OK, due to cluttering if the `ToEndpointResponse` method (see image in the orig issue)
- Same applies for dev productivity -
   - the ToEndpointResponse method has Generic parameters making it hard to write and confusing sight and intellisense.
   - Testing might be harder as Mapper cannot be mocked easility
- Performance - the `ToEndpointResponse` method uses dynamic, which limits compile time optimizations
- Compile time safety - the approach introduces support of Union Type return handling with `Results`, which is great and declares all return options of the method (https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-7.0#resultstresult1-tresultn). We need to keep this!

### Details
- Mapping - FastEndpoints comes with Mappers being integrated into the Endpoint class public abstract class Endpoint<TRequest, TResponse, TMapper> : Endpoint<TRequest, TResponse>, IHasMapper<TMapper>. We need to leverage the generic for introducing IMappers, which are recommended to map Entity objects to DTOs. This way, we can even “auto-map” the Entity to DTO using the framework and all we will need to do is to add the TMapper class in the endpoint class declaration - https://fast-endpoints.com/docs/domain-entity-mapping#mapping-logic-in-a-separate-class
- Serialization - the current of implementation of the ToEndpointResponse method doesn’t leverage the built-in serialization infrastructure available in FastEndpoints, thus the LongToStringConverter doesn’t seem to work. If you check the Implementation of the SendSync and other framework methods, you will see that they all use the Definition.SerializerContext behind the hood, e.g. return HttpContext.Response.SendAsync(response, statusCode, Definition.SerializerContext, cancellation); https://github.com/FastEndpoints/FastEndpoints/blob/b10b9473aa1f2124e7bda8a0c499ae15b705eb23/Src/Library/Endpoint/Endpoint.Send.cs#L13, while the TypedResult and ToResultAsync don’t do that and simply send the response to the context. Hence the issues with serialization.
  - => the intuition is that we should rewrite the Result mapping logic to map to each of the available framework Send methods - https://fast-endpoints.com/docs/misc-conveniences#send-methods instead of using the base Minimal API IResult based output
- Extension Method - we currently encapsulate the logic as extension method of Result, e.g.
   - Inherit from Endpoint and override ExecuteAsync or HandleAsync to implement what we need. We can also introduce a method. Example here - https://github.com/ardalis/Result/issues/175
   - Extension method of IEndpoint e.g.
- In order to have type safety for the possible returned statues, we utilize https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpresults.results-2?view=aspnetcore-8.0 to https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpresults.results-6?view=aspnetcore-8.0. The objects are created from TypedResults.Xxx (e.g. TypedResults.Created) having the assignment operator being implicitly overloaded.
- Implementing the processing logic in an extension of Result makes it complicated to have the cast. This leads to two main issues:
   - The above mentioned code not being readable because of complicated generics passed
   - The usage of dynamic keyword
- Solving the above two issues will probably be achieved with the mentioned approach to inherit Endpoint, instead of using Result extension (as described).
- I am not sure if the inheritance approach will validate the result type if we are using SendAsync (or similar). This needs to be confirmed.
```
public static class EndpointExtensions
{
    public static Task SendCode(this IEndpoint ep, int statusCode, CancellationToken ct = default)
    {
        ep.HttpContext.MarkResponseStart(); //don't forget to always do this
        ep.HttpContext.Response.StatusCode = statusCode;
        return ep.HttpContext.Response.StartAsync(ct);
    }
}
```
Recommendation: I will try with extension method of the IEndpoint first and then add custom class that inherits from Endpoint

PeterMarinov:
# Improve support for Results Union Type and Result<T> in Fast Endpoints

## Description
Adds TypedResultsBuilder to map `IResult` to `Results<TResult1, TResultN>` union types and allow for better dev experience via:
* Minimize Generic type parameters being passed on Fast Endpoint's `ExecuteAsync` method
* Compile time safety or the FastEndpoints's TResponse where TResponse is Http Union Types - https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-8.0#resultstresult1-tresultn
* Usage of implicit user-defined conversions for TypedResultsBuilder -> Results with less code and compile time safety
* Support for `IResult<TSource>` -> `IResult<TDestination>` mapping between IResult to allow better map domain entities to DTOs
* Added new `FormMapperExtensions` to show example of better mapping code for dev readability
* Added mapping tests

>[!Note]
> Note only the Form lists endpoint was touched so we can discuss. Will address all other endpoints on approval

>[!Note]
>In general implicit conversions are guaranteed to succeed and should not result in data loss (✅ ) or should not throw exceptions (❌ ) - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators. In our case we are forced to throw an `InvalidCastException` exception as this will be really unexpected case. Also it's the only way to allow us to return different types given same number and type of input arguments, so it's an Overload of the return type that works well with the C# Compiler.

```c#
 public static implicit operator Results<Ok<TData>, NotFound>(TypedResultsBuilder<TData, TResult1, TResult2> resultMapper)
    {
        return resultMapper.SourceResult.Status switch
        {
            ResultStatus.Ok => TypedResults.Ok(resultMapper.SourceResult.Value),
            ResultStatus.NotFound => TypedResults.NotFound(),
            _ => throw new InvalidCastException($"Cannot cast the {resultMapper.SourceResult.ValueType} with status {resultMapper.SourceResult.Status}")
        };
    }
```

## Example Usage With Mapper

```c#
 return TypedResultsBuilder
            .MapResult(result, forms => forms.ToFormModel())
            .ConfigureResults<Ok<IEnumerable<FormModel>>, BadRequest>();
```

## Example Usage Without Mapper

```c#
 return TypedResultsBuilder
            .FromResult(result)
            .ConfigureResults<Ok<IEnumerable<FormModel>>, BadRequest>();
```

## Related Issues
Closes #31 

## Type of Change

## Checklist


src/Endatix.Api/Infrastructure/TypedResultsBuilder.cs
```diff
@@ -0,0 +1,132 @@
...
+        SourceResult = sourceResult;
+    }
+
+    /// <summary>
+    /// Produces a new TypedResultsBuilder instance that can map to additional result types.
+    /// </summary>
+    /// <typeparam name="TResult1">The first result type to map to.</typeparam>
+    /// <typeparam name="TResult2">The second result type to map to.</typeparam>
+    /// <returns>A new instance of TypedResultsBuilder that can map to TResult1 and TResult2.</returns>
+    public TypedResultsBuilder<TData, TResult1, TResult2> ConfigureResults<TResult1, TResult2>()
```
dspetrov, line 76: As discussed a better name would be `SetTypedResults`.

PeterMarinov: PR ready to be reviewed again. Added the new task for future improvement - https://github.com/endatix/endatix/issues/168

dspetrov (approved):

PR #141, opened 2024-10-17
feat: POC AI form creation

### LINKED ISSUE #140: [POC] Endpoint for Form Creator AI Assistant
## Overview
Create a Proof of Concept for endpoint for Form Creator AI Assistant.

## Details
The endpoint should be POST assistant/forms/define with the following schema.

Request:
{
  "prompt": "string",
  "definition": "optional string",
  "assistantId": "optional string",
  "threadId": "optional string"
}

Response:
{
  "definition": "string",
  "assistantId": "string",
  "threadId": "string"
}

OpenAI Assistant API should be used. The response contains the JSON generated definition and the assistant and thread IDs that can be used to continue the adjusting the form definition while keeping the context of the conversation with the AI Assistant.

The endpoint should work for three cases:
- Only `prompt` is send in the request - A new SurveyJS form definition is created and returned based on the request. This creates a new OpenAI assistant and thread and returns their IDs.
- `prompt`, `assistantId` and `threadId` are sent in the request - An already existing thread continues to be used and a changed form definition JSON is returned, along with the assistant and thread IDs.
- `prompt` and `definition` are sent in the request - This creates a new OpenAI assistant and thread but adjusts the provided form definition according to the prompt. The response contains the the assistant and thread IDs.

## Proposed Solution
Use https://github.com/openai/openai-dotnet to access OpenAI API and the classes from OpenAI.Assistants namespace.

## Deliverables
The described endpoint and logic.

dspetrov:
# POC AI form creation

## Description
Added endpoint POST assistant/forms/define according to the description of #140

## Related Issues
closes #140

## Type of Change
- [x] POC (non-breaking change which adds functionality)

## Testing
Example testing of the three scenarios from the issue:
1. Send request with prompt="I need a survey for movie feedback", check the returned definition and keep the assistant and thread IDs. This can be the case when a new form is created from scratch with AI.
2. Send request with prompt="Add a gender question" and the assistant and thread IDs from point 1, check the returned updated definition and keep it. This can be the case to continue an ongoing chat for editing a form with AI, no matter if it's a new or existing form.
3. Send request with prompt="Add three standard questions about Terminator 2" and the returned definition from point 2, check the returned updated definition. This can be the case when an existing form is edited with AI.

For checking the returned definition https://www.toolsoverflow.com/string/unescape-string can be used for escaping characters and then https://surveyjs.io/survey-creator/examples/free-nps-survey-template/reactjs can be used to set the JSON in the tab `JSON Editor` and then see the result in the tab `Preview'

### DIFF
--- a/src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
+++ b/src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
@@ -36,12 +36,16 @@ public class DefineForm(IMediator _mediator) : Endpoint<DefineFormRequest, Resul
     public override async Task<Results<Ok<DefineFormResponse>, BadRequest>> ExecuteAsync(DefineFormRequest request, CancellationToken cancellationToken)
     {
         var result = await _mediator.Send(
-            new DefineFormCommand(request.Prompt, request.Definition),
+            new DefineFormCommand(request.Prompt, request.Definition, request.AssistantId, request.ThreadId),
             cancellationToken);
 
         return result.ToEndpointResponse<
             Results<Ok<DefineFormResponse>, BadRequest>,
-            string,
-            DefineFormResponse>(definition => new DefineFormResponse { Definition = definition });
+            AssistedDefinitionDto,
+            DefineFormResponse>(assistedDefinition => new DefineFormResponse {
+                Definition = assistedDefinition.Definition,
+                AssistantId = assistedDefinition.AssistantId,
+                ThreadId = assistedDefinition.ThreadId
+            });
     }
 }
--- a/src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
+++ b/src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
@@ -1,0 +1,47 @@
+using FastEndpoints;
+using MediatR;
+using Microsoft.AspNetCore.Http.HttpResults;
+using Endatix.Api.Infrastructure;
+using Endatix.Infrastructure.Identity.Authorization;
+using Endatix.Core.UseCases.Assistant.DefineForm;
+
+namespace Endatix.Api.Endpoints.Assistant;
+
+/// <summary>
+/// Endpoint for defining a form using AI assistance.
+/// </summary>
+public class DefineForm(IMediator _mediator) : Endpoint<DefineFormRequest, Results<Ok<DefineFormResponse>, BadRequest>>
+{
+    /// <summary>
+    /// Configures the endpoint settings.
+    /// </summary>
+    public override void Configure()
+    {
+        Post("assistant/forms/define");
+        Permissions(Allow.AllowAll);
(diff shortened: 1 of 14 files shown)


src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
```diff
@@ -0,0 +1,51 @@
...
+using Endatix.Api.Infrastructure;
+using Endatix.Infrastructure.Identity.Authorization;
+using Endatix.Core.UseCases.Assistant.DefineForm;
+
+namespace Endatix.Api.Endpoints.Assistant;
+
+/// <summary>
+/// Endpoint for defining a form using AI assistance.
+/// </summary>
+public class DefineForm(IMediator _mediator) : Endpoint<DefineFormRequest, Results<Ok<DefineFormResponse>, BadRequest>>
```
PeterMarinov, line 13: Primary constructors inject mutable parameters, so we should not use underscore prefix like we do this with private readonly fields

I think, this is now added to the text-editor config. Please change

src/Endatix.Infrastructure/Assistant/AssistantService.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        WaitToComplete(threadRun);
+        var newDefinition = await GetLastMessage(threadRun.ThreadId);
+        return new AssistedDefinitionDto(newDefinition, threadRun.AssistantId, threadRun.ThreadId);
+    }
+
+    private void WaitToComplete(ThreadRun threadRun)
+    {
+        do
+        {
```
PeterMarinov, line 67: Do we need to add try/catch to log in case of failure? I am not familiar with the OpenAI API recommendations, but wondering on your thoughts

PeterMarinov (approved):

src/Endatix.Api/Endpoints/Assistant/DefineForm.cs
dspetrov: Yes, this is how they work. I don't think we have agreed on the underscore usage as a prefix. Personally, I don't like it for any fields but we have it in almost all of the endpoints currently. I just follow what we have.
Let's decide our naming conventions and make the change everywhere, not having it inconsistent.

src/Endatix.Infrastructure/Assistant/AssistantService.cs
dspetrov: This code is based on examples provided by OpenAI. If an exception happens in `GetRun`, the situation is like in the other calls of `client`, e.g. `CreateRun`, etc - the endpoint returns 500 and the general logging is applied.
This is a POC and expectedly it doesn't handle all the details considering validations, error handling and so on. If we proceed with adding the implementation as a part of the production code, some more useful information could be logged depending on what the OpenAI API provides.

PR #164, opened 2024-10-21
feat: [POC] Return AssistantResponse in DefineForm endpoint

dspetrov:
# Return AssistantResponse in DefineForm endpoint

## Description
Added a new field in the response called `AssistantResponse` that returns a human readable response from the AI Assistant. The response's `Definition` field is now optional as not every prompt makes changes to the definition.

### DIFF
--- a/src/Endatix.Api/Endpoints/Assistant/DefineForm.DefineFormResponse.cs
+++ b/src/Endatix.Api/Endpoints/Assistant/DefineForm.DefineFormResponse.cs
@@ -6,17 +6,22 @@ namespace Endatix.Api.Endpoints.Assistant;
 public class DefineFormResponse
 {
     /// <summary>
+    /// The human-friendly response from the AI assistant.
+    /// </summary>
+    public string AssistantResponse { get; set; } = string.Empty;
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


PeterMarinov (approved):
I am approving. 

@dspetrov  - have you looked at structured responses:
- https://www.youtube.com/watch?v=6e_oFG4JVg8
- https://openai.com/index/introducing-structured-outputs-in-the-api/

If not, let's add a task to implement it as part of our future roadmap. Looks exactly what we need

dspetrov:
Yes, using structured responses is the way it worked initially with ChatCompletion API (as in the examples).

Now we use Assistants API in order to support chat continuation with chat threads. Unfortunately providing the JSON schema stopped working with Assistants API with an error "Invalid parameter: 'response_format' of type 'json_schema' is not supported with model version `gpt-4o`."

This is a bit strange because exactly the same schema used to work with 4o model before but let's not forget Assistants API is experimental so it could have certain issues with specific models.

We now achieve the same result in practice as the response format is set as JSON object, having the response description inside. We only lack the defined schema.

In the long term it will be good to have it but with the current available options, this is the way to achieve all we wish.

PR #263, opened 2024-12-05
feat: Cover Forms, FormDefinitions and Submissions endpoints with tests

dspetrov:
# Cover Forms, FormDefinitions and Submissions endpoints with tests


tests/Endatix.Api.Tests/Endpoints/Forms/UpdateTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        var request = new UpdateFormRequest
+        {
+            FormId = formId,
+            Name = "Updated Form",
+            Description = "Updated Description",
+            IsEnabled = true
+        };
+        var result = Result.Invalid();
+        
+        _mediator.Send(Arg.Any<UpdateFormCommand>(), Arg.Any<CancellationToken>())
```
PeterMarinov, line 36: No action required. Just a problem I didn't expect to happen.

Here you should be able to just pass the `request` instead of `Arg.Any<UpdateFormCommand>()` since the request is value object. 

However, even that we use a record type this doesn't work. That's weird and I will investigate at some point

tests/Endatix.Api.Tests/Endpoints/Submissions/GetByIdTests.cs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_ValidRequest_ReturnsOkWithSubmission()
+    {
+        // Arrange
+        var formId = 1L;
+        var submissionId = 1L;
+        var request = new GetByIdRequest { FormId = formId, SubmissionId = submissionId };
+        var submission = new Submission("{ }", 1, 2) { Id = submissionId };
```
PeterMarinov, line 69: Sharing thoughts.

We need some domain testing infrastructure - set of test Ids, set of test JSON definitions and responses for quick db or test seeding. 

We should try to make tests easier to write and read by humans so less things to worry about the better

Ideally, this should be shared across all projects in the test folder

PeterMarinov (approved):
@dspetrov  - this is approved. I have ran the tests few times and added my comments for aligning on general direction.

Please remember to add your prompt in a Readme in the test folder when you have time

Overall, it's great to have increased coverage even at this stage. I expect a lot of refactoring issues once we add Paging, Filtering and ProblemDetails, but let's say that's a first world problem in the world of software engineering.

tests/Endatix.Api.Tests/Endpoints/Submissions/GetByIdTests.cs
dspetrov: Generally, this is considered good practice but before going in this direction, we need to consider how it will affect the AI assistance we use for tests generation. I don't know if this is the case, but it is possible that generating/writing tests could become more time consuming if we need to use seed data.

PeterMarinov: It shouldn't matter. It should help us actually. AI will predict the use of sample data and factory/data seeding methods once we start using the. Copilot and Cursor index the codebase, so Thai will help both humans and agents

PR #123, opened 2024-09-26
feat: Share forms publicly and collect submissions with Endatix Hub

orilux:
# Share forms publicly and collect submissions with Endatix Hub

## Description
Introduces a public page for sharing forms so that they can be distributed to collect submissions. The design follows the [wireframes](https://www.figma.com/design/2bwMXMuvAOJKYXemu5HjhX/Endatix-Hub---Wireframes?node-id=0-1&node-type=canvas&t=wNT5Pb2boChRN0gM-0)

## Related Issues
Closes #122.

## Type of Change
Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots
not needed

Issue #237, opened 2024-11-26
Change JavaScript client to fire partial Submission on event


PeterMarinov:
### Details
* Logic should be in separate JS file(s), so it can be used as a package in the future
* Following events will be used:
   - onComplete
   - onValueChanged
   - onDynamicPanelItemValueChanged
   - onMatrixCellValueChanged
   - onCurrentPageChanged

### Scope

- [x] Add JavaScript logic to execute the Submission logic based of the events from the spec
- [x] Add ability to execute the API request in queue (avoid concurrent calls)
- [x] Ensure Submit form clears the queue and executes immediately
- [ ] #267
- [x] Port the logic into the Endatix Hub app
- [ ] ~~Port the logic into the Corp Website submit logic (or similar)~~ `out-of-scope`
- [ ] ~~Create npm package~~ `out-of-scope`

### Acceptance Criteria

- Complete Submission is saved in the DB when the SurveyJS's `onComplete` event is fired
- During onComplete submission Submit button cannot be resubmitted twice
- Partial Submission is saved in the DB for the following SurveyJS's events -  `onValueChanged`, `onDynamicPanelItemValueChanged`, `onMatrixCellValueChanged`, `onCurrentPageChanged`
- Server side errors during partial submission should not break the UI of the form
- Create/Update Submission API calls should be executed one at a time

PeterMarinov:
@orilux , @dspetrov  - looking at this task. The `onValueChanged` seems too frequent as it fires on individual question update and I suggest we don't add it by default to avoid storming app with partial updates on every single question. We have page changes and matrix questions, so this will be in pair with most of the form solutions out there.

While `onValueChanged` is great for analytics and understanding form lifecycle it's probably too "aggressive" of an update trigger for partial submission. Maybe we can enable with form configuration or even skip it for v1. Let me know

dspetrov:
> @orilux , @dspetrov - looking at this task. The `onValueChanged` seems too frequent as it fires on individual question update and I suggest we don't add it by default to avoid storming app with partial updates on every single question. We have page changes and matrix questions, so this will be in pair with most of the form solutions out there.
> 
> While `onValueChanged` is great for analytics and understanding form lifecycle it's probably too "aggressive" of an update trigger for partial submission. Maybe we can enable with form configuration or even skip it for v1.
> 
> Vote with 👍 or 👎 to mark your preference

[...]

PR #268, opened 2024-12-06
fix: Adding fix for current page set to zero as starting index

PeterMarinov:
# Adding fix for current page set to zero as starting index

## Description
- As part of the #237 task, have encountered a bug in the Partial submission server side validation of the CurrentPage prop. This is a fix to allow for currentPage starting from 0 as default for multi page and single page forms
- Adding couple of tests to cover related fail scenarios

## Related Issues
List any related issues (e.g., fixes #123, addresses #456).

## Type of Change

## Checklist
- [x] I have followed the accepted guideline


dspetrov (approved):
LGTM

PR #272, opened 2024-12-08
docs: Simplify the create new feature template

dspetrov:
# Simplify the create new feature template

## Description
The current create new feature template has too many sections expecting details that are not applicable for every feature. This PR simplifies it. focusing on the description of what is needed, how it will be done and what are the expected results.


PeterMarinov (approved):
Nice. Less is small

PR #273, opened 2024-12-09
fix: Redirect to Login on Session timeout

### LINKED ISSUE #193: [BUG] Forms List Page: Expired JWT Token doesn't redirect you to Login page
**Describe the bug**
When users' JWT token expires (the one issued from Endatix.API) and user hits the Forms list page, they see an error the fetching fails, but the real reason is that their JWT has expired

**To Reproduce**
1. Authenticate and go to Forms page
2. Wait for the JWT token to expire
3. Refresh page

**Expected behavior**
Until we implement the Refresh token logic, when JWT token has expired, user should be redirected to the login page to reauthenticate. Once refresh token is supported, we should extend the user session behind the scene

**Actual behavior**
User is logged out, but the following error screen is seen (check screenshot)

**Screenshots**
Image

**Desktop (please complete the following information):**
N/A

PeterMarinov:
# Redirect to Login on Session timeout

## Description
- Increased timeout from 15 to 45 minutes to increase timeout duration
- Added auth middleware to allow targeting all next JS pages
- Excluded from the middleware match rule are Next.js static folders, robots.txt, sitemaps, public/asset images, login page and server actions (as recommended)
- Changed all assets in `public` folder to be under `public/assets` to allow targeting them in the Next.js middleware and excluded them from auth middleware
- Added `ensureAuthenticated` method in the authService to allow for redirect on server-actions

>[!NOTE]
>For the actions we need to further improve the approach and use intersection and parallel rules for the Sheets instead if actions. This will allow for better redirect flow.

## Related Issues
closes #193 

## Type of Change

## Checklist
- [x] I have followed the contribution guideline and rules for the project


dspetrov (approved):
LGTM

Issue #289, opened 2024-12-10
[FEATURE] Slack notification message integration


orilux:
## **Description**
When a new form has been submitted a notification message needs to be posted into a specified Slack channel. The Slack message should contain a link to the submission details screen with the Endatix Hub.

### Slack Docs
A Slack App will need to be registered: https://api.slack.com/apps. The webhooks documentation is available here: https://api.slack.com/messaging/webhooks
### Quick implementation
The first part of the webhook article suggests that the webhook address can be generated manually. This flow can be phase 1 of the project.
The notification message should be static.
### Long-term implementation
The webhook URLs can be created programmatically (https://api.slack.com/messaging/webhooks#incoming_webhooks_programmatic)
The notification message is dynamic and contains placeholders. For example "New submission for {FormName} on {SubmittedOn}."
### Settings
- Store the notification message string in the settings
- Store the manually configured webhook address in the settings (not sure that is necessary)
### Workflow
1. Form submitter clicks "Complete".
2. Message is posted in the specified Slack channel
## **Tasks**
- [ ] Config infrastructure (API keys, etc.)
- [ ] Create Slack HTTP client
- [ ] Implement Slack notification in event handler

## **Acceptance Criteria**
- [ ] Slack App is created
- [ ] A webhook address can be configured manually (auto-generation is a nice-to-have)
- [ ] New form submissions trigger an automatic message in Slack

PR #294, opened 2024-12-11
feat: feature slack notification message integration

orilux:
# Slack integration

## Description
Created a basic client for the Slack API. Added a handler for the submission complete event. Created settings infrastructure for the Slack integration - configuration is doen through the appsettings.json file.

To test this you will need a Slack bearer token and access to a test Slack channel within the Endatix space.

The Slack message does not refer to the form by Name, due to an issue with DI of the Forms repository. A new issue will be created for it.

TODOs:

- Fix Form repo issue
- Add throttling (Slack limit is 1 message per second per channel)
- Add retry mechanism

## Related Issues
Related to issue #289 

## Type of Change

## Checklist
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

## Screenshots
image

### DIFF
--- a/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
+++ b/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
@@ -1,0 +1,12 @@
+namespace Endatix.Core.Integrations.Slack;
+
+public interface SlackCLient {
+    /// <summary>
+    /// Posts a message to a Slack channel through the Endatix Bot Slack app
+    /// </summary>
+    /// <param name="token">The Slack API bearer token</param>
+    /// <param name="channelId">The id of the channel to post the message to</param>
+    /// <param name="message">The body of the message</param>
+    /// <returns>Task</returns>
+    Task PostMessageAsync(string token, string channelId, string message);
+}
--- a/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
+++ b/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
@@ -1,6 +1,6 @@
 namespace Endatix.Core.Integrations.Slack;
 
-public interface SlackCLient {
+public interface ISlackClient {
     /// <summary>
     /// Posts a message to a Slack channel through the Endatix Bot Slack app
     /// </summary>
--- a/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
+++ b/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
@@ -1,10 +1,13 @@
+using Endatix.Core.Events;
+
 namespace Endatix.Core.Integrations.Slack;
 
 public interface ISlackClient {
     /// <summary>
     /// Posts a message to a Slack channel through the Endatix Bot Slack app
     /// </summary>
-    /// <param name="message">The body of the message</param>
+    /// <param name="notification">The notification object passed by the event publisher</param>
+    /// /// <param name="cancellationToken">The event's cancellation token</param>
     /// <returns>Task</returns>
-    private Task PostMessageAsync(string message);
+    public Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken);
 }
--- a/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
+++ b/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
@@ -9,5 +9,5 @@ public interface ISlackClient {
     /// <param name="notification">The notification object passed by the event publisher</param>
     /// /// <param name="cancellationToken">The event's cancellation token</param>
     /// <returns>Task</returns>
-    public Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken);
+    Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken);
 }
--- a/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
+++ b/src/Endatix.Core/Integrations/Slack/ISlackClient.cs
@@ -4,9 +4,7 @@ public interface ISlackClient {
     /// <summary>
     /// Posts a message to a Slack channel through the Endatix Bot Slack app
     /// </summary>
-    /// <param name="token">The Slack API bearer token</param>
-    /// <param name="channelId">The id of the channel to post the message to</param>
     /// <param name="message">The body of the message</param>
     /// <returns>Task</returns>
-    Task PostMessageAsync(string token, string channelId, string message);
+    private Task PostMessageAsync(string message);
 }
--- a/src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
+++ b/src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
@@ -1,16 +1,18 @@
 ﻿
 using Endatix.Core.Entities;
+using Endatix.Core.Events;
 using Endatix.Core.Infrastructure.Domain;
 using Endatix.Core.Infrastructure.Messaging;
 using Endatix.Core.Infrastructure.Result;
 using Endatix.Core.Specifications;
+using MediatR;
 
 namespace Endatix.Core.UseCases.Submissions.PartialUpdate;
 
 /// <summary>
 /// Handler for partially updating a form submission.
 /// </summary>
-public class PartialUpdateSubmissionHandler(IRepository<Submission> repository) : ICommandHandler<PartialUpdateSubmissionCommand, Result<Submission>>
+public class PartialUpdateSubmissionHandler(IRepository<Submission> repository, IMediator mediator) : ICommandHandler<PartialUpdateSubmissionCommand, Result<Submission>>
 {
     private const int DEFAULT_CURRENT_PAGE = 1;
 
@@ -34,6 +36,10 @@ public class PartialUpdateSubmissionHandler(IRepository<Submission> repository) 
         );
 
         await repository.SaveChangesAsync(cancellationToken);
+
+        if(submission.IsComplete) {
+            await mediator.Publish(new SubmissionCompletedEvent(submission), cancellationToken);
+        }
 
         return Result.Success(submission);
     }
--- a/src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
+++ b/src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
@@ -18,6 +18,7 @@ public class PartialUpdateSubmissionHandler(IRepository<Submission> repository, 
 
     public async Task<Result<Submission>> Handle(PartialUpdateSubmissionCommand request, CancellationToken cancellationToken)
     {
+        var mustPublishEvent = false;
         var submissionSpec = new SubmissionByFormIdAndSubmissionIdSpec(request.FormId, request.SubmissionId);
         var submission = await repository.SingleOrDefaultAsync(submissionSpec, cancellationToken);
         if (submission == null)
@@ -25,6 +26,10 @@ public class PartialUpdateSubmissionHandler(IRepository<Submission> repository, 
             return Result.NotFound("Form submission not found.");
         }
 
+        if(!submission.IsComplete && (request.IsComplete ?? false)) {
+            mustPublishEvent = true;
+        }
+        
         // TODO: add more advanced PATCH-ing where we can not only replace individual properties, but merge, remove and other typical operations. This is valid especially for the JSON based JsonData and Metadata properties, so we can keep payloads and client logic light, e.g. submit one answer at a time and update JsonData
         // TODO: investigate if IsComplete and CurrentPage should be auto calculated as part of processing the submission
         submission.Update(
@@ -37,7 +42,7 @@ public class PartialUpdateSubmissionHandler(IRepository<Submission> repository, 
 
         await repository.SaveChangesAsync(cancellationToken);
 
-        if(submission.IsComplete) {
+        if(mustPublishEvent) {
             await mediator.Publish(new SubmissionCompletedEvent(submission), cancellationToken);
         }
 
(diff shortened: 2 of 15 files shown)


src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -0,0 +1,68 @@
...
+    public SlackClient(ILogger<SlackClient> logger, IRepository<Form> formRepository, IOptions<SlackSettings> options) {
+        _formRepository = formRepository;
+        _logger = logger;
+        _slackSettings = options.Value;
+    }
+
+    public async Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken) {
+        if(notification.Submission.IsComplete && _slackSettings.Active){
+
+            var submissionUrl = $"{_slackSettings.EndatixHubBaseUrl.TrimEnd('\\', '/')}/forms/submissions/{notification.Submission.FormId}";
```
dspetrov, line 31: A good practice we follow is to not have magic strings in the code. Please use a string constant.

One option is to define a template like e.g. `private const string SubmissionUrlTemplate = "{0}/forms/submissions/{1}"` but also just defining the string literal part as constant should be fine.

Same for the the other text messages or URLs, like `message` and in `httpClient.PostAsync`.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -0,0 +1,68 @@
...
+    private readonly IRepository<Form> _formRepository;
+    private readonly SlackSettings _slackSettings;
+    public SlackClient(ILogger<SlackClient> logger, IRepository<Form> formRepository, IOptions<SlackSettings> options) {
+        _formRepository = formRepository;
+        _logger = logger;
+        _slackSettings = options.Value;
+    }
+
+    public async Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken) {
+        if(notification.Submission.IsComplete && _slackSettings.Active){
```
dspetrov, line 29: Not something necessarily needed but a suggestion. Having a not completed submission on SubmissionCompletedEvent is not an expected situation. Logging a warning in such case would bring attention if it happens.

dspetrov (changes requested):
Looks good. I have some comments about small changes and suggestions but nothing major.

src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
```diff
@@ -35,6 +37,10 @@ public async Task<Result<Submission>> Handle(PartialUpdateSubmissionCommand requ
 
         await repository.SaveChangesAsync(cancellationToken);
 
+        if(submission.IsComplete) {
```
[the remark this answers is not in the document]
orilux: Made the necessary changes - event will fire only when IsComplete was changed from false to true.

src/Endatix.WebHost/appsettings.json
```diff
@@ -46,6 +46,16 @@
       "WelcomeEmailTemplateId": "d-111e609ec2d04a8c8347cee3a759638b"
     }
   },
+  "Integrations": {
```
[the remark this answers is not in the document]
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -0,0 +1,68 @@
...
+/// </summary>
+public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlackClient
+{
+    private readonly ILogger<SlackClient> _logger;
+    private readonly IRepository<Form> _formRepository;
+    private readonly SlackSettings _slackSettings;
+    public SlackClient(ILogger<SlackClient> logger, IRepository<Form> formRepository, IOptions<SlackSettings> options) {
+        _formRepository = formRepository;
+        _logger = logger;
+        _slackSettings = options.Value;
```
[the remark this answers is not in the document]
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done with template constants. Added TODO comment to refactor this so that it is configurable through the UI

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -19,28 +21,64 @@ public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlac
     private readonly ILogger<SlackClient> _logger;
     private readonly IRepository<Form> _formRepository;
```
dspetrov, line 22: It seems we don't need the repo here and in the constructor as we're not injecting it during the object creation.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -19,28 +21,64 @@ public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlac
...
         _logger = logger;
         _slackSettings = options.Value;
+        _serviceProvider = serviceProvider;
     }
 
-    public async Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken) {
-        if(notification.Submission.IsComplete && _slackSettings.Active){
+    public async Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken)
+    {
+        Guard.Against.Null(_slackSettings.Active, nameof(_slackSettings.Active), "Slack settings must specify whether the integration is active.");
```
dspetrov, line 46: It's good that you added checks for the settings.
Something that I wish to suggest is that these checks are done in the beginning of the constructor. This way we would ensure on successful creation of the SlackClient object, it has everything needed to do its job. In addition the checks would not be necessary in the different methods of the class.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -19,28 +21,64 @@ public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlac
...
+        {
+            if (!notification.Submission.IsComplete)
+            {
+                _logger.LogWarning($"SubmissionCompletedEvent raised on an incomplete submission with id {notification.Submission.Id}");
+            }
+
+        }
     }
 
     public async Task PostMessageAsync(string token, string channelId, string message)
```
dspetrov, line 79: I have one more suggestion here. This method looks like a good candidate to become private. It seems it's only for internal use in the class (correct me if I'm wrong).
Something else I realize now - `token` and `channelId` are passed to the method but they are actually directly accessible from here as coming from a class member - `_slackSettings.Token` and `_slackSettings.ChannelId`. They can be removed from the member's input parameters.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -19,28 +21,64 @@ public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlac
...
+            {
+                var repository = scope.ServiceProvider.GetRequiredService<IRepository<Form>>();
+                form = await repository.GetByIdAsync(notification.Submission.FormId, cancellationToken);
+            }
 
-            var submissionUrl = $"{_slackSettings.EndatixHubBaseUrl.TrimEnd('\\', '/')}/forms/submissions/{notification.Submission.FormId}";
-            
-            //TODO: Troubleshoot _formRepository being disposed below and uncomment the line in order to reference the form name correctly
-            // var form = await _formRepository.GetByIdAsync(notification.Submission.FormId, cancellationToken) ?? throw new Exception("Form not found by the FormId of the submission");
+            Guard.Against.Null(form, $"Unable to load submission {notification.Submission.Id}'s Form object.");
```
dspetrov, line 63: I'm thinking about how this will affect the flow. Putting a guard clause here means leads to an exception if the form is not loaded and a message not published.
Although having the form's name in the message is good, not having the message sent because of missing form's name doesn't seem like the best approach.
I suggest if the form is not loaded a warning is logged so we can investigate and the ID is used in the message. It can be something like:
- "...|New submission> for Contact Us Form" when the form is loaded
- "...|New submission> for form with ID=9328569378562" when the form is not loaded

src/Endatix.Core/UseCases/Submissions/PartialUpdate/PartialUpdateSubmissionHandler.cs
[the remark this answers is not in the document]
dspetrov: Good, thank you.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        var content = new StringContent(
+            System.Text.Json.JsonSerializer.Serialize(payload),
+            Encoding.UTF8,
+            "application/json"
+        );
+
+        try
+        {
+            var response = await httpClient.PostAsync("https://slack.com/api/chat.postMessage", content);
```
dspetrov, line 103: Looks like a good candidate that the URL is also set as a constant, like the others.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
dspetrov: Yes, good suggestion.

dspetrov (changes requested):
Thanks for addressing my previous comments and for your suggestions. I have some more comments. They are all connected with small changes but I think will make the solution better.

src/Endatix.WebHost/appsettings.json
```diff
@@ -58,7 +58,7 @@
   },
   "Endatix": {
     "Data": {
-      "ApplyMigrations": true,
+      "ApplyMigrations": false,
```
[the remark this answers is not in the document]
orilux: Corrected

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done. Unfortunatelly now we get "possible null" warnings in the Handle function.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Done

src/Endatix.Core/UseCases/Submissions/Update/UpdateSubmissionHandler.cs
```diff
@@ -34,6 +36,10 @@ public async Task<Result<Submission>> Handle(UpdateSubmissionCommand request, Ca
 
         await repository.SaveChangesAsync(cancellationToken);
 
+        if(!submission.IsComplete && (request.IsComplete ?? false)) {
```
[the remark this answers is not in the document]
orilux: Done

src/Endatix.Core/Integrations/Slack/ISlackClient.cs
```diff
@@ -1,10 +1,13 @@
...
 public interface ISlackClient {
     /// <summary>
     /// Posts a message to a Slack channel through the Endatix Bot Slack app
     /// </summary>
-    /// <param name="message">The body of the message</param>
+    /// <param name="notification">The notification object passed by the event publisher</param>
+    /// /// <param name="cancellationToken">The event's cancellation token</param>
     /// <returns>Task</returns>
-    private Task PostMessageAsync(string message);
+    public Task Handle(SubmissionCompletedEvent notification, CancellationToken cancellationToken);
```
dspetrov, line 12: A couple of things to mention, no need to be changed now, just to have in mind:
- Interface members don't need access modifiers - they are considered public by default. I even expected that explicitly using `public` will fail the build but it doesn't have problem.
- This method returns Task so it's async method. A good practice is to end their names with `Async`, like e.g. `HandleAsync`. We don't follow this everywhere in our code base but it's good to try following it.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
dspetrov: Yes, there is a relatively new operator for such cases - putting ! after an object of nullable type explicitly says the value is not null, like e.g. `(bool)_slackSettings.Active!`. The warning disappears this way,

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
dspetrov: It's good that you introduce such fallback logic but it seems it won't really cover the case. It seems `formNameOrId` is set in the case of available form name but is never used. There is no logic that sets the ID in case there is no form and I see a guard clause that fill fail the method execution if there is no form loaded. Please revise.

dspetrov (changes requested):
Thank you for the changes, now it's definitely better. I have three new comments - both are just suggestions but one needs a small revision of the logic.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Applied

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
orilux: Thanks. Fixed.

src/Endatix.Core/Integrations/Slack/ISlackClient.cs
dspetrov: I agree. As I mentioned, we're not quite consistent on the naming of async methods. It would be good to sync about this in the team. Thanks.

dspetrov (approved):
LGTM. Seems we're ready - approving

PR #302, opened 2024-12-14
feat: Create Form sheet changes

### LINKED ISSUE #284: [FEATURE] Update Create Form Sheet according to the latest design
## **Description**
Currently the Create Form sheet does not allow to create a form with any of the four options. The user should be able to create a form from scratch for now. Follow the design below, hide the unnecessary elements (AI chat, etc) and make it possible to create a form from scratch.

create-form-sheet

## **Implementation Notes**
Sync the current implementation with the one from the design.
[Figma designs](https://www.figma.com/design/ZUK1rY2v5lah0Aorn7hivU/Endatix-High-Fidelity-Wireframes?node-id=0-1&node-type=canvas&t=207S3Q2KJTMt5Z8o-0)

## **Acceptance Criteria**
- [x] The four buttons on the Create Form sheet are displayed as in the design
- [x] The user can proceed to the Form Design screen when choosing to design from scratch
- [x] The other three buttons do not initiate the form creation
- [x] The AI form creation elements are not displayed

dspetrov:
# Create Form sheet changes

## Description
Changes on Create Form sheet following the acceptance criteria in the task and the design.

## Related Issues
closes #284

## Type of Change

## Checklist

### DIFF
--- a/apps/endatix-hub/app/(main)/forms/create-form.action.ts
+++ b/apps/endatix-hub/app/(main)/forms/create-form.action.ts
@@ -1,0 +1,27 @@
+'use server'
+
+import { ensureAuthenticated } from "@/lib/auth-service";
+import { CreateFormRequest, CreateFormResult } from "@/lib/form-types";
+import { createForm } from "@/services/api";
+
+export async function createFormAction(request: CreateFormRequest): Promise<CreateFormResult> {
+  await ensureAuthenticated();
+
+  const result: CreateFormResult = {
+    isSuccess: false
+  }
+
+  try {
+    const form = await createForm(request);
+    if (form.id?.length > 0) {
+      result.formId = form.id;
+      result.isSuccess = true;
+    } else {
+      result.error = "Failed to create form";
+    }
+  } catch (error) {
+    console.error("Failed to create form", error);
+  } finally {
+    return result;
+  }
+}
--- a/apps/endatix-hub/app/(main)/forms/ui/create-form-sheet.tsx
+++ b/apps/endatix-hub/app/(main)/forms/ui/create-form-sheet.tsx
@@ -8,15 +8,19 @@ import {
   SheetHeader,
   SheetTitle,
 } from "@/components/ui/sheet";
-import { FC, useState } from "react";
+import { FC, useState, useTransition } from "react";
 import ChatBox from "./chat-box";
-import { Atom, BicepsFlexed, Code, Folder } from "lucide-react";
+import { Atom, BicepsFlexed, Code, Copy, Folder } from "lucide-react";
 import { cn } from "@/lib/utils";
 import DotLoader from "@/components/loaders/dot-loader";
 import { showComingSoonMessage } from "@/components/layout-ui/teasers/coming-soon-link";
+import { CreateFormRequest } from "@/lib/form-types";
+import { createFormAction } from "../create-form.action";
+import { redirect } from "next/navigation";
 
 type CreateFormOption =
   | "from_scratch"
+  | "from_existing"
   | "from_template"
   | "from_json"
   | "via_assistant";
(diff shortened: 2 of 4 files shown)


apps/endatix-hub/app/(main)/forms/create-form.action.ts
```diff
@@ -0,0 +1,27 @@
+'use server'
+
+import { ensureAuthenticated } from "@/lib/auth-service";
+import { CreateFormRequest, CreateFormResult } from "@/lib/form-types";
+import { createForm } from "@/services/api";
+
+export async function createFormAction(request: CreateFormRequest): Promise<CreateFormResult> {
+  await ensureAuthenticated();
+
+  const result: CreateFormResult = {
```
PeterMarinov, line 10: I have added Result type, for the idea of Action results. Can you try to use it. It's in the `submit-form.action.ts`. You can see it has helpers similar to the Result we use in the C#. 

We can merge as it is, but I would like if you can try with it to get first hand experience and provide feedback as we evolve it.

P.S. From the sample in the `submit-form.action`, the exporting of type that wraps the generic is optional. I just find it nice :

`export type SubmissionOperationResult = Result<SubmissionOperation>; //This is optional`

PeterMarinov (approved):
@dspetrov  - I have approved. Please check my comment to see if the Result type will help you. It's new one, so hoping you can use it. Should take you just few minutes if it's OK. If not forget it

apps/endatix-hub/app/(main)/forms/create-form.action.ts
dspetrov: Thanks for pointing this, I didn't know we have it.
I've made a change - you can see it in the last commit.
Generally it didn't need much changes and the resulting code seems nice and tidy. Two things I need to mention that I find could be better or I couldn't figure out.
- Checking the status happens with a method call, like e.g. `Result.isSuccess(formResult)`. Wouldn't it be better if we have something like `formResult.isSuccess`?
- I couldn't find how to access the error message the I set in the result object so I just don't read it from the result.

P.S. The PR now needs another approval.

apps/endatix-hub/app/(main)/forms/ui/create-form-sheet.tsx
```diff
@@ -80,10 +81,11 @@ const CreateFormSheet = () => {
             formDefinitionJsonData: JSON.stringify("{ }")
         }
         const formResult = await createFormAction(request);
-        if (formResult.isSuccess && formResult.formId) {
-            redirect(`/forms/${formResult.formId}`);
+        if (Result.isSuccess(formResult) && formResult.value) {
```
PeterMarinov, line 84: Here idea is you can just say `Result.isSuccess()`, but I see that's not guaranteed yet. I need to change the API to make the value check redundant.

apps/endatix-hub/app/(main)/forms/ui/create-form-sheet.tsx
```diff
@@ -80,10 +81,11 @@ const CreateFormSheet = () => {
...
+            const formId = formResult.value;
+            redirect(`/forms/${formId}`);
         } else {
-            alert(formResult.error);
```
PeterMarinov: I see what the problem is. Error only has message and errorType as required props. If you don't see them in the auto-complete, then let me know. 

```type Error = {
    kind: Kind.Error;
    errorType: ErrorType;
    message: string;
    details?: string;
};
```

For the future, I will change the API to return decomposed object with the possible states, e.g.:

```ts
const [success, error] = createFormResult();
if (error){
//return early with error
}

console.log(success.value); // happy path
```

PeterMarinov (approved):
Let's merge. I have added comments how the Result type can be improved for the future. It look fairly easy to do. Check comments before you merge (no changes required)

apps/endatix-hub/app/(main)/forms/ui/create-form-sheet.tsx
dspetrov: I couldn't see something about error. Now I think maybe it should be message but strangely it doesn't appear in the auto-complete.

Issue #230, opened 2024-11-19
Change Form PATCH and POST update form methods to generate new version on JSON Schema changes


PeterMarinov:
### Goals
Now that we have schema, we must handle the creation of new Form Definitions according to the new requirements

### Scope
- [ ] Ensure Form and Form Definition update API calls generate new Active Definition only of there are submissions associated with 
- [ ] Add Unit Tests to ensure Command Handlers and API endpoints behave as expected

### Acceptance Criteria
- When an ActiveForm definition is updated and there are NO submissions associated with the Definition, the Active Definition is updated and NO new Definition is created
- When an ActiveForm definition has its JSON DATA UPDATED and there are EXISTING submissions associated with the Definition, the NEW Definition is created and it is set as an Active Definition
- When an ActiveForm definition is updated, but its JSON DATA IS NOT UPDATED,  the Active Definition is updated and NO new Definition is created

PR #304, opened 2024-12-15
feat: change form patch and post update form methods to generate new version on json schema changes

orilux:
# Form definition versioning when existing submissions are present

## Description

- When an ActiveForm definition is updated and there are NO submissions associated with the Definition, the Active Definition is updated and NO new Definition is created
- When an ActiveForm definition has its JSON DATA UPDATED and there are EXISTING submissions associated with the Definition, the NEW Definition is created and it is set as an Active Definition
- When an ActiveForm definition is updated, but its JSON DATA IS NOT UPDATED, the Active Definition is updated and NO new Definition is created

## Related Issues
#230 

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Core/Specifications/SubmissionsTotalCountByFormDefinitionIdSpec.cs
+++ b/src/Endatix.Core/Specifications/SubmissionsTotalCountByFormDefinitionIdSpec.cs
@@ -1,0 +1,23 @@
+using Ardalis.Specification;
+using Endatix.Core.Entities;
+using Endatix.Core.Specifications.Common;
+using Endatix.Core.Specifications.Parameters;
+
+namespace Endatix.Core.Specifications;
+
+
+/// <summary>
+/// Returns the total count of submissions per form.
+/// </summary>
+public class SubmissionsTotalCountByFormDefinitionIdSpec : Specification<Submission>
+{
+    /// <summary>
+    /// Initializes a new instance of the specification to retrieve submissions for a given form definition
+    /// </summary>
+    /// <param name="formDefinitionId">The ID of the form definition to retrieve submissions for</param>
+    public SubmissionsTotalCountByFormDefinitionIdSpec(long formDefinitionId)
+    {
+        Query.Where(s => s.FormDefinition.Id == formDefinitionId)
+            .AsNoTracking();
+    }
+}
--- a/src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
+++ b/src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
@@ -1,12 +1,13 @@
 ﻿using Endatix.Core.Abstractions.Repositories;
 using Endatix.Core.Entities;
+using Endatix.Core.Infrastructure.Domain;
 using Endatix.Core.Infrastructure.Messaging;
 using Endatix.Core.Infrastructure.Result;
 using Endatix.Core.Specifications;
 
 namespace Endatix.Core.UseCases.FormDefinitions.PartialUpdateActive;
 
-public class PartialUpdateActiveFormDefinitionHandler(IFormsRepository formRepository) : ICommandHandler<PartialUpdateActiveFormDefinitionCommand, Result<FormDefinition>>
+public class PartialUpdateActiveFormDefinitionHandler(IFormsRepository formRepository, IRepository<Submission> submissionsRepository) : ICommandHandler<PartialUpdateActiveFormDefinitionCommand, Result<FormDefinition>>
 {
     public async Task<Result<FormDefinition>> Handle(PartialUpdateActiveFormDefinitionCommand request, CancellationToken cancellationToken)
     {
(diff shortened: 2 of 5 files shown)


src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
```diff
@@ -18,10 +19,24 @@ public async Task<Result<FormDefinition>> Handle(PartialUpdateActiveFormDefiniti
...
+            activeDefinition.UpdateDraftStatus(request.IsDraft);
+        }
 
         await formRepository.UpdateAsync(formWithActiveDefinition, cancellationToken);
         return Result.Success(activeDefinition);
     }
+
+    private async Task<bool> FormDefinitionHasSubmissions(long definitionId) {
+        var spec = new SubmissionsTotalCountByFormDefinitionIdSpec(definitionId);
+        var count = await submissionsRepository.CountAsync(spec);
```
PeterMarinov, line 39: Can you rewrite this with `AnyAsync()`. It is recommended for performance when we don't need the count, but if there are items at all - https://admirmujkic.medium.com/why-the-any-method-can-be-faster-than-count-in-c-9921a092a492

src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
```diff
@@ -18,10 +19,24 @@ public async Task<Result<FormDefinition>> Handle(PartialUpdateActiveFormDefiniti
             return Result.NotFound("Active form definition not found.");
         }
 
-        activeDefinition.UpdateSchema(request.JsonData);
-        activeDefinition.UpdateDraftStatus(request.IsDraft);
+        if(await FormDefinitionHasSubmissions(activeDefinition.Id) && activeDefinition.JsonData != request.JsonData) {
```
PeterMarinov, line 22: I would recommend to reverse the if statements with the cheapest first and then go to the database. It's good practice to add the most expensive check (e.g. DB round trip) at the end and only check it when other if conditions lead to it.

Also, I would recommend extract this if check into a private method for readability e.g. `Task<bool> ShouldCreateNewFormDefinitionAsync()` will make the code easier to debug and read.

src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
```diff
@@ -18,10 +19,24 @@ public async Task<Result<FormDefinition>> Handle(PartialUpdateActiveFormDefiniti
...
+            var newFormDefinition = new FormDefinition(jsonData: request.JsonData);
+            await formRepository.AddNewFormDefinitionAsync(formWithActiveDefinition, newFormDefinition, cancellationToken);
```
PeterMarinov, line 24: I am wondering if we need this method at all? Can we just add the FormDefinition using the Form entity, set it as Active and then call the `await formRepository.UpdateAsync(formWithActiveDefinition, cancellationToken);` e.g. don't do this SaveChanges?

Will be great if it works since FormDefinitions are tracked by the parent Form entity. On the dark side, there might be some EF exceptions.... You don't have to do anything before the merge, but wondering if you have tried 

In both cases, I would like us to try this when we have a bit more time and see if we can handle the whole thing in one SaveChanges call. 

P.S. I have checked the `CreateFormDefinitionHandler` and it looks it will work in which case you can delete this repository method and just go with this approach using the Form as your aggregate root and update the FormDefinition through it:

```c#
var newFormDefinition = new FormDefinition(request.IsDraft, request.JsonData);
        form.AddFormDefinition(newFormDefinition);
        if (!newFormDefinition.IsDraft)
        {
            form.SetActiveFormDefinition(newFormDefinition);
        }
        await formsRepository.UpdateAsync(form, cancellationToken);
```

dspetrov (approved):
Looks good. I've commented something very small that looks a bit confusing in a name.
Something else that came to my mind and is not urgent but can create problems in the future if we forget it - it's good that we have the same logic not only in `PartialUpdateActiveFormDefinitionHandler` but also in `UpdateActiveFormDefinitionHandler` as it can be used to pass all the fields for update. The two handlers are generally quite similar. If you think it's a quick and easy change, please go for it. Otherwise, let's create a task for this.

PeterMarinov (changes requested):
Oggy, this look nice. 

I have added some comments to improve readability and hopefully make the code tinier and easy for the reader

src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
```diff
@@ -1,27 +1,49 @@
...
-        activeDefinition.UpdateDraftStatus(request.IsDraft);
+        if(await ShouldCreateNewFormDefinitionAsync(activeDefinition, request)) {
+            var newFormDefinition = new FormDefinition(request.IsDraft??false, request.JsonData);
+            formWithActiveDefinition.AddFormDefinition(newFormDefinition);
+            
+            if (!newFormDefinition.IsDraft) {
+                formWithActiveDefinition.SetActiveFormDefinition(newFormDefinition);
+            }
+           
+            await formsRepository.UpdateAsync(formWithActiveDefinition, cancellationToken);
```
PeterMarinov, line 30: I see that this line will invoke alongside the other `UpdateAsync` on line 37. Is this intentional or probably can be removed?

PeterMarinov (approved):
Approved

src/Endatix.Core/UseCases/FormDefinitions/PartialUpdateActive/PartialUpdateActiveFormDefinitionHandler.cs
PeterMarinov: I like that you have been able to remove the repository method

src/Endatix.Core/Specifications/SubmissionsTotalCountByFormDefinitionIdSpec.cs
```diff
@@ -0,0 +1,23 @@
...
+/// <summary>
+/// Returns the total count of submissions per form.
+/// </summary>
+public class SubmissionsTotalCountByFormDefinitionIdSpec : Specification<Submission>
+{
+    /// <summary>
+    /// Initializes a new instance of the specification to retrieve submissions for a given form definition
+    /// </summary>
+    /// <param name="formDefinitionId">The ID of the form definition to retrieve submissions for</param>
+    public SubmissionsTotalCountByFormDefinitionIdSpec(long formDefinitionId)
```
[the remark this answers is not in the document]
PeterMarinov: Suggested `SubmissionsTotalCountByFormDefinitionIdSpec` => `SubmissionsByFormDefinitionIdSpec`

dspetrov (approved):

PR #307, opened 2024-12-15
feat: Design Form saving behavior changes

### LINKED ISSUE #286: [FEATURE] Design Form saving behavior changes
## **Description**
The current saving behavior on the Form Design screen is unclear. The save needs to be explicit from a button, while the the back button only warns for unsaved changes before going back but does not save. An informative label should show if there are unsaved changes.

design-form

## **Implementation Notes**
The state if there are unsaved changes should be kept and a label "unsaved changes" should be added (it is not present in the current design). The back button should not save the form but directly go back or ask with the browser "Stay on page" dialog in case there are changes.
[Figma designs](https://www.figma.com/design/ZUK1rY2v5lah0Aorn7hivU/Endatix-High-Fidelity-Wireframes?node-id=0-1&node-type=canvas&t=207S3Q2KJTMt5Z8o-0)

## **Acceptance Criteria**
- [x] Do not save on back button
- [x] On back button click show the browser "Stay on page" dialog if there are changes
- [x] Show a label "unsaved changes" if there are unsaved changes
- [x] Save the changes on click of Save button

dspetrov:
# Design Form saving behavior changes

## Description
Includes the described in #286 changes and additionally on an attempt to close the browser with unsaved changes there is a confirmation dialog.

**Note**
The user still can leave the screen with unsaved changes if they click to some other screen on the side menu bar. Confirmation dialog for this case is not a part of the task. However, I tried to add such logic but it seems changes need to be made on higher level in the whole application to make this possible so I left it as a future enhancement.

## Related Issues
closes #286

## Type of Change

## Checklist

### DIFF
--- a/apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
+++ b/apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
@@ -125,7 +125,7 @@ const defaultCreatorOptions: ICreatorOptions = {
 
 function FormEditor({ formJson, formId, formName, options }: FormEditorProps) {
   const [creator, setCreator] = useState<SurveyCreator | null>(null);
-  const [isSaving, setIsSaving] = useState(false);
+  const [isSaving] = useState(false);
   const router = useRouter();
   const [isEditingName, setIsEditingName] = useState(false);
   const [name, setName] = useState(formName);
--- a/apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
+++ b/apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
@@ -271,7 +271,7 @@ function FormEditor({ formJson, formId, formName, options }: FormEditorProps) {
         </div>
         <div className="flex items-center gap-2">
           {hasUnsavedChanges && (
-            <span className="text-yellow-600 text-xs border border-yellow-600 px-2 py-0.5 rounded-full whitespace-nowrap">
+            <span className="font-bold text-black text-xs border border-black px-2 py-0.5 rounded-full whitespace-nowrap">
               Unsaved changes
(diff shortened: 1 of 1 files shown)


apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
```diff
@@ -255,7 +269,12 @@ function FormEditor({ formJson, formId, formName, options }: FormEditorProps) {
             </span>
           )}
         </div>
-        <div className="flex items-center gap-2" >
+        <div className="flex items-center gap-2">
+          {hasUnsavedChanges && (
+            <span className="text-yellow-600 text-xs border border-yellow-600 px-2 py-0.5 rounded-full whitespace-nowrap">
```
Comment (orilux) [inline apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx:274]: Yellow on a white background is not the most contrastful combination. I would go for green and make the text bold, or maybe even try to make the background green anf the text white.

orilux (approved):

apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx
Comment (dspetrov) [reply apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx:274]: You're right, thanks for the suggestion. I'll make changes but I only think is green the right for this? We want to warn the user and green doesn't sound to me really for warning. What do you think?

Comment (orilux) [reply apps/endatix-hub/app/(main)/forms/[formId]/ui/form-editor.tsx:274]: Yellow doesn't work on white, unless you make the whole background yellow and make the text black. Red is wrong, because it suggests an error. Green is not ideal for the reasons you stated, but it grabs attention better, because everything else we have in color is in blue.
Most sites we looked at for reference don't actually have any color for this status label. It is black and white. I believe that's the case with Github too.

orilux (approved):

PR #315, opened 2024-12-16
feat: Do not return submissions JSON data in List Submissions endpoint

### LINKED ISSUE #313: [FEATURE] Do not return submissions JSON data in List Submissions endpoint
## **Description**
Do not return  data in List Submissions endpoint does not need the submissions JSON data as it is not displayed in the grid and is just making the responses much larger.

## **Implementation Notes**
Consider the option to not even read the JSON from the DB.

## **Acceptance Criteria**
- [x] Do not return submissions JSON data in List Submissions endpoint

dspetrov:
# Do not return submissions JSON data in List Submissions endpoint

## Description
A temp solution to not return submissions JSON data in List Submissions endpoint.

**Things that need to be addressed**
- A new constructor is added in `Submission` to receive `id`, `createdAt` and `completedAt` because the `Select` call in `SubmissionsByFormIdSpec` needs to create an object and set these values. Consider a better solution.
- The endpoint uses `SubmissionModel`, which is used by `GetById` endpoint and now `ListByFormId` returns some of the fields with null/empty values. Separate models should be used that have only the needed fields.

## Related Issues
closes #313

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Core/Entities/Submission.cs
+++ b/src/Endatix.Core/Entities/Submission.cs
@@ -9,7 +9,6 @@ public partial class Submission : BaseEntity, IAggregateRoot
 
     public Submission(string jsonData, long formId, long formDefinitionId, bool isComplete = true, int currentPage = 1, string? metadata = null)
     {
-        Guard.Against.NullOrEmpty(jsonData, nameof(jsonData));
         Guard.Against.NegativeOrZero(formId, nameof(formId));
         Guard.Against.NegativeOrZero(formDefinitionId, nameof(formDefinitionId));
 
@@ -20,6 +19,15 @@ public partial class Submission : BaseEntity, IAggregateRoot
         Metadata = metadata;
 
         SetCompletionStatus(isComplete);
+    }
+
+    public Submission(long id, string jsonData, long formId, long formDefinitionId, bool isComplete, int currentPage, string? metadata, 
+        DateTime createdAt, DateTime? completedAt) 
+        : this(jsonData, formId, formDefinitionId, isComplete, currentPage, metadata)
+    {
+        Id = id;
+        CreatedAt = createdAt;
+        CompletedAt = completedAt;
     }
 
     public bool IsComplete { get; private set; }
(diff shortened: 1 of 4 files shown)


src/Endatix.Core/Specifications/SubmissionsByFormIdSpec.cs
```diff
@@ -12,7 +12,7 @@ namespace Endatix.Core.Specifications;
 /// TODO: [check] We can add PagedSpecification&lt;T&gt; instead of Specification&lt;T&gt; as one direction to reuse and encapsulate logic, but we need to factor in other requirements like Basic filtering/sorting/ordering + specific filtering, e.g. only ActiveForms
 /// TODO: [check] Also handle Ability to return PagedResult, which has the current page number and total count of items instead of basic list of results
 /// </summary>
-public class SubmissionsByFormIdSpec : Specification<Submission>
+public class SubmissionsByFormIdSpec : Specification<Submission, Submission>
```
PeterMarinov, line 15: My recommendation is to change the class definition from
`public class SubmissionsByFormIdSpec : Specification<Submission, Submission>`

to
`public class SubmissionModelsByFormIdSpec : Specification<Submission, SubmissionModel>`

Idea is to add to the domain logic that the list requirements only need part of the Submission. Maybe as we move to a vertically sliced approach where all the layers will be in the same assembly that will be more natural too

src/Endatix.Core/Entities/Submission.cs
```diff
@@ -22,6 +21,15 @@ public Submission(string jsonData, long formId, long formDefinitionId, bool isCo
         SetCompletionStatus(isComplete);
     }
 
+    public Submission(long id, string jsonData, long formId, long formDefinitionId, bool isComplete, int currentPage, string? metadata,
```
PeterMarinov, line 24: I am for merging this fast if needed, but please do check my recommendation for using specification with entity projection to a DTO.

PeterMarinov (dismissed):
Approved with single comment on suggestions

src/Endatix.Core/Entities/Submission.cs
dspetrov: Now I use SubmissionDto and it works without the need of changes in Submission entity class. Thanks for the suggestion.
We need to clarify what types of objects (Entity objects, API models, DTOs) the different endpoints return because now it's inconsistent.

Issue #325, opened 2024-12-17
[FEATURE] Auto-resize images


orilux:
## **Description**
Implement automatic image resizing to reduce payload size.

## **Implementation Notes**

- Ensure the file is an image before attempting to resize it.
- Use "canvas" (https://imagekit.io/blog/how-to-resize-image-in-javascript/)
- Store the maximum dimensions in the config

## **Acceptance Criteria**

- [ ] Images are reduced in size upon upload

PR #330, opened 2024-12-17
feat: Image resizing

orilux:
# Image resizing

## Description
Intercepts uploaded images and scales them down to a smaller file size to optimize payload size.

### Configuration
Environment variable `NEXT_PUBLIC_MAX_IMAGE_SIZE` accepts numeric values only. A value of 0 disables auto-resizing. A value larger than 0 indicates the maximum allowed length of the longer side of an uploaded image. Images are always resized proportionally.

### How it works

It handles the [onValueChanging](https://surveyjs.io/form-library/documentation/api-reference/survey-data-model#onValueChanging) SurveyJS event. Since file upload question may accept multiple files, the script iterates through the array of files passed by the event. Every file's type is checked to see whether it is an image and then it is processed.

A check is made to see which side of the image is larger. If the larger side of the image is larger than the `NEXT_PUBLIC_MAX_IMAGE_SIZE` value, then the image is resized.

### Known issues

This script strips EXIF information from the file.

## How to test

After experimenting with a large image, the resulting BASE64 string may be rendered at this website: https://base64.guru/converter/decode/image

## Related Issues
#325 

## Type of Change

## Checklist


apps/endatix-hub/features/public-form/ui/survey-component.tsx
```diff
@@ -58,7 +60,57 @@ export default function SurveyComponent({ definition, formId, submission }: Surv
     });
   }, []);
 
+  const interceptImages = useCallback((sender: SurveyModel, options: ValueChangingEvent) => {
```
PeterMarinov, line 63: Have you considered renaming the `options` parameter to `event`? Since it's an event, event might be a clearer name and make it easier to understand the data being passed. Let me know what you think

apps/endatix-hub/features/public-form/ui/survey-component.tsx
```diff
@@ -58,7 +60,57 @@ export default function SurveyComponent({ definition, formId, submission }: Surv
...
+        img.src = imgSrc;
+      });
+    };
+    
+    const processImages = async () => {
+    
+      for (let i = 0; i < options.question.value.length; i++) {
+        if (options.value[i].type.includes('image')) {
+            options.value[i].content = await resizeImage(
+              options.value[i].content,
```
PeterMarinov, line 102: Having a local variable will be faster and easier to read instead of going through the array reference every time. Can you refacto to extract a variable for `options.value[i]`

apps/endatix-hub/features/public-form/ui/survey-component.tsx
```diff
@@ -58,7 +60,57 @@ export default function SurveyComponent({ definition, formId, submission }: Surv
...
+          }
+        };
+        img.src = imgSrc;
+      });
+    };
+    
+    const processImages = async () => {
+    
+      for (let i = 0; i < options.question.value.length; i++) {
+        if (options.value[i].type.includes('image')) {
```
PeterMarinov, line 100: Are you sure that the `.type` property will be available for every question? It looks to me that's an assumption only true for questions with files. Please add safety to avoid null reference.

Also I would recommend to rename this if check to a boolean to add meaning to it. For example `isValueAnImage` or smth like that. Will make the code read better. Let me know what you think

apps/endatix-hub/features/public-form/ui/survey-component.tsx
```diff
@@ -58,7 +60,57 @@ export default function SurveyComponent({ definition, formId, submission }: Surv
...
 
+  const interceptImages = useCallback((sender: SurveyModel, options: ValueChangingEvent) => {
+    const resizeImage = (imgSrc: string, type: string): Promise<string> => {
+      return new Promise((resolve) => {
+        const img = new Image();
+        img.onload = () => {
+          let width = img.width;
+          let height = img.height;
+    
+          if (width > height) {
```
PeterMarinov, line 71: I want to recommend a simplification to make this more expressive and easier to understand. What do you think of this:

```ts
if (width > height && width > MAX_IMAGE_SIZE) {
  // handle this case
} else if (height > MAX_IMAGE_SIZE) {
 // handle this case
}
```

PeterMarinov (changes requested):
Great to have the solution in place. I have suggested few minor refactors to improve readability and one, which I think it's a bug with the `.type` accessor.

Also as we discussed, we should aim to extract various features into separate files to keep the survey component clean and easy to maintain. This way, we can even tree shake not needed features based of use case and deliver smaller bundle size. You don't have to do it in this PR, but I will try to add the pattern before this code becomes too big and torublesome

apps/endatix-hub/.env.example
```diff
@@ -23,3 +23,6 @@ SLACK_CLIENT_SECRET=
 
 # [OPTIONAL] Slack redirect uri. Optional if you don't want to use Slack.
 SLACK_REDIRECT_URI=
+
+# The number of pixels to resize the longer side of an image to. A value of 0 disables automatic image resizing.
+NEXT_PUBLIC_MAX_IMAGE_SIZE=0
```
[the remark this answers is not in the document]
orilux: Done

apps/endatix-hub/features/public-form/ui/survey-component.tsx
orilux: Done

apps/endatix-hub/features/public-form/ui/survey-component.tsx
orilux: Done

apps/endatix-hub/features/public-form/ui/survey-component.tsx
PeterMarinov: Thanks. I am aware that the "options" is standard naming convention for this parameter across events. However, since this represents our custom logic built on top of it, my recommendation was to ensure the code better aligns with our context

I don’t consider this a blocking issue for the PR, so I’m happy to leave it as is for now.

apps/endatix-hub/features/public-form/ui/survey-component.tsx
```diff
@@ -97,16 +93,17 @@ export default function SurveyComponent({ definition, formId, submission }: Surv
...
+          value.content = await resizeImage(
+            value.content,
+            value.type
             );
         }
       }
     };
     
-    if(MAX_IMAGE_SIZE > 0) processImages();
+    if(MAX_IMAGE_SIZE > 0 && options.question.getType() == "file") processImages();
```
PeterMarinov, line 106: This will solve the issue with null reference. I want to recommend couple of things (small ones)

1. Add null safety operator to `value.type.includes('image')` => `value?.type?.includes('image')` This way code will be safe
2. For debugging we try to avoid one line ifs and we would convert the 
`if(MAX_IMAGE_SIZE > 0 && options.question.getType() == "file") processImages();`
to 
```
if(MAX_IMAGE_SIZE > 0 && options.question.getType() == "file") {
   processImages();
}

We are consistent with this across both c# and ts

PeterMarinov (approved):
@orilux  - Approved with cosmetic changes comment. I double checked the codes and it looks to me the code resizes already resized images. I might be wrong about it, but can you check if we don't resize images multiple times.

Feel free to merge. I have approved

apps/endatix-hub/features/public-form/ui/survey-component.tsx
orilux: Done

PeterMarinov (approved):

PR #372, opened 2025-01-17
feat: Internal test forms filtering and Slack message bypass

orilux:
# A temporary workaround for filtering internal test forms

## Description

Applies a global query filter for all forms with a CreatedAt date that is earlier than 2021-01-01. This filter is ignored for user accounts with @endatix.com email address.

Slack messages are skipped for all forms with CreatedAt date that is earlier than 2021-01-01.

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
+++ b/src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
@@ -69,7 +69,7 @@ public class SlackClient : INotificationHandler<SubmissionCompletedEvent>, ISlac
 
             // TODO: Remove when per-account settings are implemented
             if(form != null && form.CreatedAt.CompareTo(new DateTime(2021,1,1)) < 0) {
-                _logger.LogWarning($"Functional test form - skip Slack notification.");
+                _logger.LogInformation($"Functional test form - skip Slack notification.");
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


src/Endatix.WebHost/Program.cs
```diff
@@ -3,6 +3,7 @@
 var builder = WebApplication.CreateBuilder(args);
 
 builder.Services.AddHealthChecks();
+builder.Services.AddHttpContextAccessor();
```
PeterMarinov, line 6: I am afraid, that's not a good place to add this service registration as it will mix the role of the WebHost, which should be just the bare minimum to run the app.

Please move the `AddHttpContextAccessor()` registration to the `AddInfrastructure` method as part of the `src/Endatix.Infrastructure/Setup/EndatixAppExtensions.cs`.

src/Endatix.Infrastructure/Integrations/Slack/SlackClient.cs
```diff
@@ -67,6 +67,12 @@ public async Task Handle(SubmissionCompletedEvent notification, CancellationToke
                 form = await repository.GetByIdAsync(notification.Submission.FormId, cancellationToken);
             }
 
+            // TODO: Remove when per-account settings are implemented
+            if(form != null && form.CreatedAt.CompareTo(new DateTime(2021,1,1)) < 0) {
+                _logger.LogWarning($"Functional test form - skip Slack notification.");
```
PeterMarinov, line 72: @orilux  - please change this to `LogInformation` as it will show as an error. I think, this should be Information level based of other logs in the system

PeterMarinov (changes requested):

src/Endatix.WebHost/Program.cs
orilux: Done

PeterMarinov (approved):
Approved. Ready for merge

PR #373, opened 2025-01-17
fix: Set DateTime kind to UTC in filters

dspetrov:
# Set DateTime kind to UTC in filters

## Description
When having time in API filters, the kind should be set in the DateTime objects for Postgres queries to work properly.

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Core/Specifications/Common/SpecificationHelper.cs
+++ b/src/Endatix.Core/Specifications/Common/SpecificationHelper.cs
@@ -92,6 +92,14 @@ public static class SpecificationHelper
             return Enum.Parse(targetType, value, true);
         }
 
+        if (targetType == typeof(DateTime) || targetType == typeof(DateTime?))
+        {
+            var dateTime = DateTime.Parse(value);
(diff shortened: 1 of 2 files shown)


src/Endatix.Core/Specifications/Common/SpecificationHelper.cs
```diff
@@ -92,6 +92,14 @@ private static object ParseValue(string value, Type targetType)
             return Enum.Parse(targetType, value, true);
         }
 
+        if (targetType == typeof(DateTime) || targetType == typeof(DateTime?))
```
PeterMarinov, line 95: just a node. With pattern matching this can be represented as `targetType is typeof(DateTime) or typeof(DateTime?)`

PeterMarinov (approved):
Approved. Just FYI comment

src/Endatix.Core/Specifications/Common/SpecificationHelper.cs
PeterMarinov: Haven't tested though... but wanted to share

PR #377, opened 2025-01-25
feat: Support form deletion

### LINKED ISSUE #362: [FEATURE] Delete form
## **Description**
Users should be able to delete forms from the Hub. 

## **Implementation Notes and considerations**

- Consider implementing a soft-delete flag.
- Consider adding a DeletedAt date-time field.
- If implementing with a soft-delete flag, consider [EF Core global query filters](https://learn.microsoft.com/en-us/ef/core/querying/filters) for more reliable filtering across all CRUD endpoints.

When deleting a form, its child objects should be cascade-deleted as well:

- Form definitions
- Submissions
- Uploaded blobs

A clear warning should be displayed before deletion.

## **Tasks**

- [ ] Warning and confirmation dialog UI mockup
- [x] Schema changes
- [x] Query filters
- [x] Endpoints
- [x] UI implementation

dspetrov:
# Support form deletion

## Description
DB schema changes, Endatix API and Hub implementation of form deletion support.
Use the Delete menu item in the Form Sheet dots menu. Deleting a from deletes all its definitions and submissions. Deletion is soft deletion setting a new column `IsDeleted` in every entity table.

## Related Issues
closes #362

## Type of Change

## Checklist

### DIFF
--- a/apps/endatix-hub/features/forms/ui/form-sheet.tsx
+++ b/apps/endatix-hub/features/forms/ui/form-sheet.tsx
@@ -1,7 +1,7 @@
 'use client'
 
 import { Button } from "@/components/ui/button"
-import { Copy, Link2, List, Pencil } from 'lucide-react';
+import { Copy, Link2, List, Pencil, MoreHorizontal, Trash2, AlertTriangle } from 'lucide-react';
 import {
     Sheet,
     SheetContent,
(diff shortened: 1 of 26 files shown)


src/Endatix.Persistence.PostgreSql/Setup/EndatixAppExtensions.cs
```diff
@@ -35,6 +35,16 @@ public static IEndatixApp AddPostgreSqlDataPersistence(this IEndatixApp endatixA
         endatixApp.Services.AddDbContext<AppDbContext>((serviceProvider, options) =>
         {
             options.UseNpgsql(connectionString, db => db.MigrationsAssembly(migrationsAssembly));
+
+            options.UseSeeding((context, cancellationToken) =>
```
PeterMarinov, line 39: @dspetrov - it looks that you will need to repeat the same for MS SQL, which includes:
- use of `UseSeeding`
- Migrations for the new columns

apps/endatix-hub/features/forms/ui/form-sheet.tsx
```diff
@@ -20,19 +20,116 @@ import { updateFormStatusAction } from "../../../app/(main)/forms/[formId]/updat
...
 
+interface DeleteFormDialogProps {
+    isOpen: boolean;
+    onOpenChange: (open: boolean) => void;
+    formName: string;
+    submissionsCount: number;
+    onDelete: () => Promise<void>;
+}
+
+const DeleteFormDialog = ({
```
PeterMarinov, line 56: I don't see this as a must do for this PR, but it's certainly worth sharing directions. We must try to put each new feature into it's own feature part for 2 reasons.

1. Keep this `form-sheet.tsx` lean. Right now it is starting to violate SR and is getting too large
2. Will keep all Delete actions & components close to each other -  higher cohesion

Addressing this now is purely up to you, but wanted to share this and potentially we can discuss on one of the upcoming arch meetings. 

I think the change-password is good example, login as well.

PeterMarinov (changes requested):
Great to have this feature. Looking good and it is more than welcome :)

2 comments. Adding support for MS SQL is the only required item.

src/Endatix.Persistence.PostgreSql/Setup/EndatixAppExtensions.cs
dspetrov: Thanks for pointing this, I missed it.
I've added the seeding and the migrations for MS SQL Server.

apps/endatix-hub/features/forms/ui/form-sheet.tsx
dspetrov: What you say makes sense for me. I've checked the change password example and thought of similar approach. However, more questions arose than answers came to my mind. The password change is more like a separate feature, while the deletion, although making sense by itself, is more like an action on the form (it just has a confirmation dialog and an API call).
I prefer we leave it the way it is for now and discuss separately our approaches for when to separate the implementations as a standalone features, so we have better separation for future changes.

PR #403, opened 2025-03-13
feat: Form Templates DB changes and endpoints

dspetrov:
# Form Templates DB changes and endpoints

## Description
A new table FormTempalates is created and CRUD endpoints are created for the operations:
- Create Form Template
- Get Form Template
- List Form Templates
- Partially Update Form Template
- Delete Form Template

## Related Issues
closes https://github.com/endatix/endatix-private/issues/126

## Type of Change

## Checklist


src/Endatix.Api/Endpoints/FormTemplates/Create.CreateFormTemplateValidator.cs
```diff
@@ -0,0 +1,29 @@
...
+/// Validation rules for the <c>CreateFormTemplateRequest</c> class.
+/// </summary>
+public class CreateFormTemplateValidator : Validator<CreateFormTemplateRequest>
+{
+    /// <summary>
+    /// Default constructor
+    /// </summary>
+    public CreateFormTemplateValidator()
+    {
+        RuleFor(x => x.Name)
```
PeterMarinov, line 17: @dspetrov  I see that the `CreateFormTemplateRequest` has the Name, IsEnabled and JsonData as optional, but here they are required. Is this the pattern we have or is something that we need to address.

Maybe it's something I forgot why it is the way it is, but I was thinking if we can make them compile time required or at least not optional as this will remove a lot of compile time warnings, if checks and etc

PeterMarinov (approved):
This looks great! 

Added one question on the use of optional when validator requires is it in Endpoint requests. Maybe something old or not, but it bugged me.

On a side node, we need to think of a pattern for light vs full response models for the endpoints and come up with consistent pattern. No issues with this PR, but the use of `FormTemplateModelWithoutJsonData` made me think we need to discuss and agree on approach. Just a topic for 📐 consideration

dspetrov:
> This looks great!
> 
> Added one question on the use of optional when validator requires is it in Endpoint requests. Maybe something old or not, but it bugged me.
> 
> On a side node, we need to think of a pattern for light vs full response models for the endpoints and come up with consistent pattern. No issues with this PR, but the use of `FormTemplateModelWithoutJsonData` made me think we need to discuss and agree on approach. Just a topic for 📐 consideration

I agree we better discuss how to have a consistent pattern for the responses. I don't really like `FormTemplateModelWithoutJsonData` but this was the easiest way to return the data without an always empty `JsonData` field for the List Form Templates endpoints (which doesn't need to return the JSON).

src/Endatix.Api/Endpoints/FormTemplates/Create.CreateFormTemplateValidator.cs
PeterMarinov: Aha... it makes sense now of why we would do it this way..

PR #407, opened 2025-03-25
fix: Create template issue

### LINKED ISSUE #406: bug: Cannot Create Template
**Describe the bug**
When creating a template the UI fails with error 

**To Reproduce**
1. Go to templates page
2. Click on "Create a Template"
3. UI fails with error. The backend reports "Tenant_Id" is negative or zero, while the user has a positive TenantId context

**Expected behavior**
Form can be created

PeterMarinov:
# fix: Create template issue

## Description
- Fixing issue with conflict between manual transaction and EF Core retry mechanism
- Changing the multitenancy middleware order to fix issue with tenant context

## Related Issues
fixes #406 

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Hosting/Builders/EndatixMiddlewareBuilder.cs
+++ b/src/Endatix.Hosting/Builders/EndatixMiddlewareBuilder.cs
@@ -46,8 +46,8 @@ public class EndatixMiddlewareBuilder
         _logger?.LogInformation("Configuring middleware with default settings");
 
         UseExceptionHandler()
-            .UseMultitenancy()
             .UseSecurity()
+            .UseMultitenancy()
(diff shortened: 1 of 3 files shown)


PeterMarinov:
> LGTM Something I'm curious - why we have this problem with Form Template creation and not with Form creation?

It's middleware related, so it will reproduce across all areas where passing the TenantId via the HttpContext is critical. In this case, was the Guard clauses of the CreateTemplateRequest model, so yeah...

dspetrov:
CreateForm also has such guard clause and I've tested it definitely - there wasn't a problem.
Is it a problem that just happens sometimes? So I was lucky when testing the Form creation?

Issue #421, opened 2025-04-23
Rename ActiveDefinitionDto to FormDefinitionView


PeterMarinov:
The class ActiveDefinitionDto (in src/Endatix.Core/UseCases/FormDefinitions/ActiveDefinitionDto.cs) is currently used for both active and inactive form definitions. To reflect its broader usage and remove the "active" implication, rename it to FormDefinitionView.

### Acceptance Criteria:

- Rename the file to FormDefinitionView.cs
- Rename the class to FormDefinitionView
- Update all references across the codebase
- Ensure naming aligns with current conventions for read models/view models

### Context:

> "Currently we use this class for active definitions but it generally keeps the data of form definition, no matter if it's active or not." — @dspetrov
> "We can rename to View. I will add a task." — @PeterMarinov

https://github.com/endatix/endatix/pull/415#discussion_r2048779588

PR #429, opened 2025-05-07
feat: Custom Questions persistence and create/list endpoints

dspetrov:
# Custom Questions persistence and create/list endpoints

## Description
- CustomQuestion domain class
- Migrations to create the CustomQuestions table and insert Video question
- Command and Query to add and list Custom Questions
- API endpoints to add and list Custom Questions

### Note
Unit tests to be added

## Related Issues
closes https://github.com/endatix/endatix-private/issues/187

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Api/Endpoints/CustomQuestions/Create/Create.cs
+++ b/src/Endatix.Api/Endpoints/CustomQuestions/Create/Create.cs
@@ -1,0 +1,44 @@
+using FastEndpoints;
+using MediatR;
+using Microsoft.AspNetCore.Http.HttpResults;
+using Endatix.Api.Infrastructure;
+using Endatix.Infrastructure.Identity.Authorization;
+using Endatix.Core.UseCases.CustomQuestions.Create;
+
+namespace Endatix.Api.Endpoints.CustomQuestions;
+
+/// <summary>
+/// Endpoint for creating a new custom question.
+/// </summary>
+public class Create(IMediator mediator) : Endpoint<CreateCustomQuestionRequest, Results<Created<CreateCustomQuestionResponse>, BadRequest>>
+{
+    /// <summary>
+    /// Configures the endpoint settings.
+    /// </summary>
+    public override void Configure()
+    {
+        Post("customization/questions");
+        Permissions(Allow.AllowAll);
+        Summary(s =>
+        {
+            s.Summary = "Create a new custom question";
+            s.Description = "Creates a new custom question with the provided data.";
+            s.Responses[201] = "Custom question created successfully.";
+            s.Responses[400] = "Invalid input data.";
+        });
+    }
+
+    /// <inheritdoc/>
+    public override async Task<Results<Created<CreateCustomQuestionResponse>, BadRequest>> ExecuteAsync(CreateCustomQuestionRequest request, CancellationToken cancellationToken)
+    {
+        var command = new CreateCustomQuestionCommand(
+            request.Name!,
+            request.JsonData!,
+            request.Description);
+        var result = await mediator.Send(command, cancellationToken);
+        
+        return TypedResultsBuilder
+            .MapResult(result, CustomQuestionMapper.Map<CreateCustomQuestionResponse>)
+            .SetTypedResults<Created<CreateCustomQuestionResponse>, BadRequest>();
+    }
+}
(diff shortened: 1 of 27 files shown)


src/Endatix.Api/Endpoints/CustomQuestions/Create/Create.cs
```diff
@@ -0,0 +1,44 @@
...
+/// Endpoint for creating a new custom question.
+/// </summary>
+public class Create(IMediator mediator) : Endpoint<CreateCustomQuestionRequest, Results<Created<CreateCustomQuestionResponse>, BadRequest>>
+{
+    /// <summary>
+    /// Configures the endpoint settings.
+    /// </summary>
+    public override void Configure()
+    {
+        Post("customization/questions");
```
PeterMarinov, line 20: I think, this name "customization" is not a good parent resource name as it's too generic. I think, we need to change it to something like "form-builder" and all other things we do remain there e.g. "form-builder/questions" or "form-builder/presets"

I am not picky, just trying to make this hard to change in the future. I am concerned with this not working for us once we start adding more customizations like workflows, integrations and other things not part of the form-builder. 

I might be wrong, of course, but let's discuss

src/Endatix.Api/Endpoints/CustomQuestions/CustomQuestionMapper.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// </summary>
+    /// <typeparam name="T">The type of the custom question API model, which inherits CustomQuestionModel.</typeparam>
+    /// <param name="question">The custom question entity.</param>
+    /// <returns>The mapped custom question API model.</returns>
+    public static T Map<T>(CustomQuestion question) where T : CustomQuestionModel, new() => new T
+    {
+        Id = question.Id.ToString(),
+        Name = question.Name,
+        Description = question.Description,
+        JsonData = question.JsonData,
```
PeterMarinov, line 21: Reminder: You wanted to move form JsonData to Model as naming. I guess, we missed the train for this one and we will do all JsonData to Model in separate PR

src/Endatix.Api/Endpoints/FormDefinitions/FormDefinitionModel.cs
PeterMarinov, line 1: Just a reminder (no action needed) -> we have this task in the backlog - [Rename ActiveDefinitionDto to FormDefinitionView](https://github.com/endatix/endatix/issues/421)
#421

src/Endatix.Core/UseCases/FormDefinitions/GetActive/GetActiveFormDefinitionHandler.cs
```diff
@@ -17,9 +22,18 @@ public async Task<Result<ActiveDefinitionDto>> Handle(GetActiveFormDefinitionQue
             return Result.NotFound("Active form definition not found.");
         }
 
+        var tenantId = formWithActiveDefinition.TenantId;
+
+        var customQuestions = await customQuestionsRepository.ListAsync(
```
PeterMarinov, line 27: Do we need this since there's EFCore QueryFilter by tenant id? Is this because this is public and the EFCore filter cannot pick the tenantId from the context?

If that's the case, I guess we might set it in the middleware for such tenant agnostic requests -> not sure what will be best way to ensure it propagates as soon as we identify it

PeterMarinov (approved):
Hey @dspetrov  - It looks great to me. I have approved, but please check my comments mostly for alignment.

src/Endatix.Api/Endpoints/CustomQuestions/Create/Create.cs
dspetrov: I absolutely agree and thanks for pointing this, I've been thinking what will be the appropriate naming and just put "customization" with the idea to discuss it later, having the other entities there like you mention (also move themes there).

Let's have e quick discussion about this as it will be much faster than writing all out points and pros/cons for the different options.

src/Endatix.Api/Endpoints/CustomQuestions/CustomQuestionMapper.cs
dspetrov: You're right and my understanding was that we'll do it separately so we won't have the different naming at once. I'll create a task about it.

PeterMarinov (approved):

Issue #170, opened 2024-10-29
Webhooks: Add Documentation and Code Sample


PeterMarinov:
## Goal
Enable customization and integration of Endatix with external solutions by providing improved documentation and practical examples for the WebHooks feature.

## Details

- The main use case is to trigger custom webhooks on Survey (Form) Create, Update, and Status Change events.
- The webhook payload should include at minimum:
     - The Form ID
     - Metadata about the change (e.g., which fields were changed, which user performed the action, and other relevant information)
- This enables triggering custom workflows, such as responding to form status changes (e.g., isEnabled/disabled).

### Scope
- Provide updated and complete sample code demonstrating:
     - How to register a custom webhook
     - How to implement a webhook handler
- Update the documentation with a dedicated "How to" guide (e.g., Guides > Using Web Hooks), similar to [Next.js Guides](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html).
- Reference and build upon:
     - [How to test webhooks locally](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
     - The implementation of the `OnSubmissionComplete` hook and handler in the samples (e.g., ContactUs form, Slack integration cases)

### Resources
- [How to test webhooks locally](https://github.com/endatix/kb/blob/main/How-Tos/test-webhooks-locally.md)
- Review the OnSubmissionComplete hook and handler in sample projects (ContactUs form, Slack integration)

PR #430, opened 2025-05-08
feat: Add form events and webhook samples

dspetrov:
# Add form events and webhook samples

## Description
Added events:
- FormCreatedEvent
- FormUpdatedEvent
- FormEnabledStateChangedEvent

Added webhook event handlers in Endatix.Samples.CustomEventHandlers project:
- FormCreatedHandler
- FormUpdatedHandler
- FormEnabledStateChangedHandler
- SubmissionCompletedHandler

## Related Issues
#170

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Core/UseCases/Forms/Update/UpdateFormHandler.cs
+++ b/src/Endatix.Core/UseCases/Forms/Update/UpdateFormHandler.cs
@@ -1,24 +1,37 @@
 ﻿using Endatix.Core.Entities;
+using Endatix.Core.Events;
 using Endatix.Core.Infrastructure.Domain;
 using Endatix.Core.Infrastructure.Messaging;
 using Endatix.Core.Infrastructure.Result;
+using MediatR;
 
 namespace Endatix.Core.UseCases.Forms.Update;
 
-public class UpdateFormHandler(IRepository<Form> _repository) : ICommandHandler<UpdateFormCommand, Result<Form>>
+public class UpdateFormHandler(
+    IRepository<Form> repository,
+    IMediator mediator) : ICommandHandler<UpdateFormCommand, Result<Form>>
 {
     public async Task<Result<Form>> Handle(UpdateFormCommand request, CancellationToken cancellationToken)
     {
-        var form = await _repository.GetByIdAsync(request.FormId, cancellationToken);
+        var form = await repository.GetByIdAsync(request.FormId, cancellationToken);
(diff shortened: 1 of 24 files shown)


src/Endatix.Core/UseCases/Forms/Update/UpdateFormHandler.cs
```diff
@@ -1,24 +1,37 @@
...
         form.Description = request.Description;
         form.IsEnabled = request.IsEnabled;
-        await _repository.UpdateAsync(form, cancellationToken);
+        await repository.UpdateAsync(form, cancellationToken);
+
+        await mediator.Publish(new FormUpdatedEvent(form), cancellationToken);
+
+        if (oldIsEnabled != request.IsEnabled)
+        {
+            await mediator.Publish(new FormEnabledStateChangedEvent(form, request.IsEnabled), cancellationToken);
```
PeterMarinov, line 32: This is the Mediator way of doing the domain even dispatching. Not sure if we should do this now, but the ideal solution is to raise the even from the domain entities and this way commands will be focused on business logic and not domain logic. Handlers will be leaner and raising events will be consistent. This will work nicely especially if the domain design enforces the behavior, e.g. you cannot change the status unless you call the `Form.Enable()`

We actually have the infrastructure to do this. All entities inherit from `HasDomainEventsBase` and as such in the Form or other entities within the mutation method, we can say `base.RegisterDomainEvent(new FormEnabledStateChangedEvent(this));`

I see you have tests that cover this so it might take you longer even though it's a simple change. If you decide not to do it, let's discuss to get your thoughts and add a task to track it

PeterMarinov (dismissed):
It's great to see the evolution of the events. Nice work.

My only comment is for the raising the events from the Domain Entities vs Handlers. I want us to discuss this and make a decision we can log and track.

I was hoping to see the docs in the same PR, so up to you to merge it or add it in a seperate one.

src/Endatix.Core/UseCases/Forms/Update/UpdateFormHandler.cs
PeterMarinov: @dspetrov  task here - [Raise Domain Events from Entities (Clean Architecture, DDD)](https://github.com/endatix/endatix/issues/431#top)
#431

PeterMarinov (approved):
Approved

PR #437, opened 2025-06-03
feat: Register User endpoint support for public account creation

dspetrov:
# Register User endpoint support for public account creation

## Description
To allow public account creation the Register User endpoint needs to be opened to public and the user should not be associated with a tenant before the email verification.

**Note:** Temporarily the endpoint is still kept not public before the implementation of email verification - https://github.com/endatix/endatix-private/issues/139

## Related Issues
https://github.com/endatix/endatix-private/issues/136

## Type of Change

## Checklist


orilux (approved):
lgtm

PR #441, opened 2025-06-25
feat: verify-email endpoint

dspetrov:
# verify-email endpoint

## Description
- Creates entity `EmailVerificationToken` and its persistance
- Creates endpoint POST `auth/verify-email`
- On user registration the new user has `EmailConfirmed=false` and `TenantId=0`
- On user registration a verification token is created and stored in the DB
- `auth/verify-email` verifies the user if the provided token is valid, not expired and not user.

Note: There are details about the verification described [here](https://github.com/endatix/endatix/blob/63ac60e11852a9a917b58fb6c422d65b133ef759/docs/email-verification.md).

## Related Issues
[[FEATURE] Create email verification endpoints](https://github.com/endatix/endatix-private/issues/145)

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Infrastructure/Identity/EmailVerification/EmailVerificationService.cs
+++ b/src/Endatix.Infrastructure/Identity/EmailVerification/EmailVerificationService.cs
@@ -1,0 +1,133 @@
+using System.Security.Cryptography;
+using Ardalis.GuardClauses;
+using Endatix.Core.Abstractions;
+using Endatix.Core.Entities.Identity;
+using Endatix.Core.Infrastructure.Domain;
+using Endatix.Core.Infrastructure.Result;
+using Endatix.Core.Specifications;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.Options;
+
+namespace Endatix.Infrastructure.Identity.EmailVerification;
+
+/// <summary>
+/// Implements the email verification service.
+/// </summary>
+public class EmailVerificationService : IEmailVerificationService
+{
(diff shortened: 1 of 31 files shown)


src/Endatix.Infrastructure/Identity/EmailVerification/EmailVerificationService.cs
```diff
@@ -0,0 +1,133 @@
...
+        
+        foreach (var existingToken in existingTokens)
+        {
+            await _tokenRepository.DeleteAsync(existingToken, cancellationToken);
+        }
+
+        // Create new token
+        var tokenValue = GenerateToken();
+        var expiresAt = DateTime.UtcNow.AddHours(_options.TokenExpiryInHours);
+        var verificationToken = new EmailVerificationToken(userId, tokenValue, expiresAt);
```
PeterMarinov, line 68: @dspetrov  - not sure if you have tried the Identity's builtin Token logic in the UserManager's `GenerateEmailConfirmationTokenAsync` - https://github.com/aspnet/AspNetIdentity/blob/main/src/Microsoft.AspNet.Identity.Core/UserManager.cs#L1141

dspetrov: I haven't but during previous tasks connected with auth (e.g. refresh token but I think other too) I've tried hardly to use build in Identity logic, which at the end couldn't work for our needs and it just turned out to waste a lot of time. Based on this experience, I decided to go for straight solution to make it faster and cleaner.

PR #460, opened 2025-08-06
feat: Authenticate webhook messages

### LINKED ISSUE #458: Authenticate webhook messages
## **Description**
Often the receivers of the webhook messages need authentication to accept the message. We need support configurable authentication.

## **Implementation Notes**
Initially we need to support adding a header to the webhook http request `X-SURVEY-API-KEY: some-api-key`. This needs to be configurable. Considering the webhooks are configured in the app settings, a good place for this configuration is there.

Next steps - putting the auth configuration in the app settings is not per tenant so this needs to be moved to tenant specific configurations.

## **Acceptance Criteria**
- [x] Webhook messages can be authenticated with http request header `X-SURVEY-API-KEY: some-api-key`
- [x] The auth details can be configured

dspetrov:
# Authenticate webhook messages

## Description
Authentication can be configured now for each webhook endpoint like
```
    "WebHooks": {
      "Events": {
        "FormCreated": {
          "IsEnabled": true,
          "WebHookEndpoints": [
            {
              "Url": "https://api.example.com/webhooks",
              "Authentication": {
                "Type": "ApiKey",
                "ApiKeyHeader": "X-SURVEY-API-KEY",
                "ApiKey": "your-secret-api-key"
              }
            }
          ]
        },
        ...
```

The old format
```
    "WebHookUrls": [
      "https://webhook.site/no-auth-endpoint"
    ]
```
is kept for compatibility and can be used in a mixed way with the new auth format.

There wasn't any test coverage of the webhook logic I tests were added for the files where there are changes.

Future enhancements will involve configuring per tenant and with more options.

## Related Issues
closes #458 

## Type of Change

## Checklist

### DIFF
--- a/docs/endatix-docs/docs/guides/webhooks.mdx
+++ b/docs/endatix-docs/docs/guides/webhooks.mdx
@@ -142,32 +142,57 @@ To test your webhooks:
       "Events": {
         "FormCreated": {
           "IsEnabled": true,
-          "WebHookUrls": [
-            "https://webhook.site/your-unique-id"
+          "WebHookEndpoints": [
+            {
+              "Url": "https://webhook.site/your-unique-id",
+              "Authentication": {
+                "Type": "None"
+              }
+            }
           ]
         },
         "FormUpdated": {
           "IsEnabled": true,
-          "WebHookUrls": [
-            "https://webhook.site/your-unique-id"
+          "WebHookEndpoints": [
+            {
+              "Url": "https://webhook.site/your-unique-id",
+              "Authentication": {
+                "Type": "None"
+              }
+            }
           ]
         },
         "FormEnabledStateChanged": {
           "IsEnabled": true,
-          "WebHookUrls": [
-            "https://webhook.site/your-unique-id"
+          "WebHookEndpoints": [
+            {
+              "Url": "https://webhook.site/your-unique-id",
+              "Authentication": {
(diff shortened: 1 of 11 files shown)


PeterMarinov (approved):
I like that there's backward compatibility and the fact that we are adding basic protection on top of the hooks. These are are valuable features that improve the product and that makes me happy. Great stuff

I am approving and we can merge. 👍 

My areas of considerations are listed below. Let's review if we should address any of the items below prior to the beta release:
- **naming** - we should consider renaming this to WebHook signatures. Webhooks are fire and forget and it's responsibility of the consumer to validate the signature to ensure origin is legitimate and not malicious. It's not a responsibility of the webhook to authenticate against the consumer. It's the other way around imo. For reference check HubSpot. They have nice article on the topic - https://developers.hubspot.com/docs/guides/apps/authentication/validating-requests
- **per endpoint authentication** - this should be per account imo. I like that it's flexible, but I think we should enforce it on the entire app/tenant with standardized settings. The mindset that if the consumer doesn't validate the signature it's their choice, but we recommend they do and we still send the data. I think it creates too much settings instead of global webhook settings per account/app
- **header-key** - I think, we should remove the option to change the header name but invest in more robust signature that is harder to intersect and replicate and provides more certainty to consumers of the legitimate origin

PR #463, opened 2025-08-07
fix: Unable to change password due to user not being able to be retrieved

dspetrov:
# Unable to change password due to user not being able to be retrieved

## Description
Loading the user from the db by principal claims was not working due to recent changes. Changed the logic to use the used ID go get the current user from the DB.

## Related Issues
closes https://github.com/endatix/endatix-private/issues/358

## Type of Change

## Checklist

### DIFF
--- a/src/Endatix.Api/Endpoints/MyAccount/ChangePassword.cs
+++ b/src/Endatix.Api/Endpoints/MyAccount/ChangePassword.cs
@@ -5,13 +5,14 @@ using Microsoft.AspNetCore.Http.HttpResults;
 using Endatix.Api.Infrastructure;
 using Endatix.Core.UseCases.MyAccount.ChangePassword;
 using Endatix.Infrastructure.Identity.Authorization;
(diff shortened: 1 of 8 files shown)


PeterMarinov (approved):
Approve - can you log in task to note that the `Task<Result<User>> GetUserAsync(ClaimsPrincipal claimsPrincipal, CancellationToken cancellationToken = default);` in the IUserService doesn't work

Maybe add a comment in the method, so we can fix

dspetrov:
> ✅ Approved - can you log in task to note that the `Task<Result<User>> GetUserAsync(ClaimsPrincipal claimsPrincipal, CancellationToken cancellationToken = default);` in the IUserService doesn't work
> 
> Maybe add a comment in the method, so we can fix

I've created the following issue:
[IUserService.GetUserAsync(ClaimsPrincipal claimsPrincipal...) doesn't work](https://github.com/endatix/endatix/issues/464)

PR #470, opened 2025-08-20
feat: Create table SubmissionVersions and store history of each JSON data change

dspetrov:
# Create table SubmissionVersions and store history of each JSON data change

## Description
Storing submissions JSON data on every modification is essential for:
- troubleshooting
- support of submission versioning in the future

The current approach provides the storage and logic to keep it but doesn't take care of how the data is used and its retention period.

## Related Issues
closes https://github.com/endatix/endatix-saas/issues/59

## Type of Change

## Checklist


src/Endatix.Infrastructure/Data/Config/SubmissionVersionConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+        builder.ToTable("SubmissionVersions");
+
+        builder.Property(sv => sv.Id)
+            .IsRequired();
+
+        builder.Property(sv => sv.SubmissionId)
+            .IsRequired();
+
+        builder.Property(sv => sv.JsonData)
+            .IsRequired();
```
PeterMarinov, line 20: Check the`ConversationConfiguration` where we have example of JSONB support for Postgres. 
`builder.Property(x => x.ResultJson).HasColumnType("jsonb");`

I see we now support only PostgreSQL. Not sure if we should add SQL Server in which case the upper comment may slow you down, so up to you to ignore if you want to add SQL Server support

src/Endatix.Infrastructure/Data/Interceptors/SubmissionVersioningInterceptor.cs
```diff
@@ -0,0 +1,81 @@
...
+        if (eventData.Context is not AppDbContext dbContext)
+        {
+            return ValueTask.FromResult(result);
+        }
+
+        CreateSubmissionVersions(dbContext);
+        return ValueTask.FromResult(result);
+    }
+
+    private void CreateSubmissionVersions(AppDbContext dbContext)
```
PeterMarinov, line 44: My main comment is to move this to the domain e.g. Submission entity or the `PartialUpdateSubmissionHandler` if that will be easier.

Will give us reasons for the code to fail and make its behavior in oru control instead of EF Core, which we can only test with integration tests

PeterMarinov (changes requested):
Great to add the change quickly!

I have added my comments, which I believe we should address as part of this PR before we merge.

src/Endatix.Core/Entities/SubmissionVersion.cs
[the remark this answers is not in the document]
dspetrov: Considering we don't use it anywhere as a standalone entity (actually we're not using it any way for now), it should be fine to go without the tenant - I've removed it.

src/Endatix.Infrastructure/Data/Interceptors/SubmissionVersioningInterceptor.cs
dspetrov: I've made these changes as discussed.

PeterMarinov (approved):
Approved. Added this task to plan proper JSON support that will unlock partial JSON patching of the submission and other benefits
[Add JSON Database Persistence Support for SQL Server and PostgreSQL](https://github.com/endatix/endatix/issues/472)
#472

PR #487, opened 2025-09-15
feat!: Base Permission System Foundation

PeterMarinov:
# feat!: Base Permission System Foundation
BREAKING CHANGE: Marks `Allow.AllowAll` as Obsolete

## Description
Creates foundational permission system infrastructure including core services, enums, and base interfaces to support granular permission-based authorization throughout the application. 

- [x] Added Actions class to define basic set of actions in the Endatix system to define the permission system.
- [x] Permission Category make as Value Object same as SubmissionStatus
- [x] Add IPermissionService and PermissionService implementation
- [x] Add HybridCache with stats on cache hits and misses
- [x] Adds Roles class to define baseline Roles definition we can use for testing and Database seeding
- [x] Add `JwtClaimsTransformer` to populate the current identiy with claims
- [x] Added test endpoint

>[!TIP]
> To test, 
>  1. Seed test data via - `https://localhost:5001/api/test/unified` and then use the test users such as 
>  2. Use Login with test user e.g. Scripter with "scripter@test.com"/ "TestPassword123!"
> 3. And last hit the UserInfo to check loaded permissions and roles - `https://localhost:5001/api/my-account/user-info`

###  TODOs
- [ ] Wire PermissionCategory to Permission entity

### Questions:
- Permission Category Questions should be generalized to Plugin or Survey Extensions e.g. expressions, calculations, etc.
- My Account - This should be "Authenticated" level, not permission-based
- No special permission needed for self-management e.g. change password, change email, etc. - AllowAnonymous() for these endpoints
- Shall we store the tenantId int he cache key or userId is sufficient?
- Review if the Cache Invaldiation cannot be seperated from the PermissionService?
- Discuss if we need hierarchical roles and permissions?
- IsAdmin - what's the logic we need to implement as well as the scope?
- `EnrichWithUserPermissionsAsync` - do we need to add a claim for "IsViewer" e.g. fast frontend acccess. Also let's review strategy to add all vs selected permission sets to the ClaimsIdentity

## Related Issues
- closes https://github.com/endatix/endatix-private/issues/264

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Screenshots
From userInfo response of the Scripter test user

```json
{
    "claims": {
        "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier": "1417082115702390784, 1417082115702390784",
        "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress": "scripter@test.com",
        "tenant_id": "1",
        "nbf": "1757929679",
        "exp": "1757983679",
        "iat": "1757929679",
        "iss": "endatix-api",
        "aud": "endatix-hub",
        "http://schemas.microsoft.com/ws/2008/06/identity/claims/role": "Scripter",
        "permission": "analytics.view, analytics.view.owned, forms.create, forms.delete.owned, forms.edit, forms.edit.owned, forms.publish, forms.publish.owned, forms.view, forms.view.owned, submissions.create, submissions.edit.owned, submissions.export, submissions.export.owned, submissions.submit, submissions.update.partial, submissions.view, submissions.view.owned, submissions.view.public.form, system.health.check, templates.create, templates.delete.owned, templates.edit, templates.edit.owned, templates.view, templates.view.owned"
    }
}
```


src/Endatix.Infrastructure/Identity/Authentication/JwtClaimsTransformer.cs
```diff
@@ -1,52 +1,174 @@
...
-        ;
 
+        return principal;
+    }
 
-        if (principal.Identity is not ClaimsIdentity primaryIdentity)
+    private static long? ExtractUserId(ClaimsPrincipal principal)
+    {
+        // Priority order: sub (JWT standard) -> NameIdentifier (ASP.NET standard) -> custom user_id (legacy)
+        var userIdClaim = principal.FindFirst(ClaimNames.UserId) ??
```
dspetrov, line 69: I think there is similar logic in user context implementation. Can't we reuse the same logic in all the places?
We'll need to make some changes in it (for "sub" but most probably no only) and following DRY will be good.

src/Endatix.Infrastructure/Identity/Authentication/JwtClaimsTransformer.cs
```diff
@@ -1,52 +1,174 @@
...
+            Permissions = roleInfo.Permissions.ToList(),
+            IsAdmin = isAdmin
+        };
+    }
+
+    /// <summary>
+    /// Gets the set of permissions that are actually used by endpoints.
+    /// This reduces memory usage by only adding claims that FastEndpoints will check.
+    /// </summary>
+    private static HashSet<string> GetUsedPermissions()
```
dspetrov, line 154: I believe this is something forgotten.

src/Endatix.Infrastructure/Identity/Services/PermissionService.cs
```diff
@@ -0,0 +1,694 @@
...
+using Endatix.Infrastructure.Identity.Authorization;
+using Microsoft.EntityFrameworkCore;
+
+namespace Endatix.Infrastructure.Identity.Services;
+
+/// <summary>
+/// High-performance permission service with multi-level caching and optimization.
+/// Uses hybrid caching, in-memory fast cache, and batch operations for enterprise performance.
+/// </summary>
+internal sealed class PermissionService : IPermissionService
```
dspetrov, line 18: It seems there is code that is never user like e.g. `GetRolePermissionsWithIncludeAsync` but I think not only, e.g. public methods not used anywhere too.
I also checked the usage of the methods in `IPermissionService`. Apart from `IsUserAdminAsync` and `GetUserRoleInfoAsync`, the others are only used internally in `PermissionService` and for testing purposes in `UnifiedTestEndpoint`.
I recommend we expose in the interface only the methods that are needed, make all other private and remove the code that is not needed. This way the implementation will be easier to understand and maintain.

dspetrov (changes requested):
Good job! I haven't gone in the details of the permissions and roles as we will surely need some adjustments in the exact actions and roles - this PR implements the foundations and this is achieved. I have a few comments connected with removing redundant code to make the implementation easier to understand and maintain.

src/Endatix.Infrastructure/Identity/Authentication/JwtClaimsTransformer.cs
PeterMarinov: I have used it. Just a note that we need a convenient override that returns the Id as long as we not need to cast it

src/Endatix.Infrastructure/Identity/Services/PermissionService.cs
PeterMarinov: Initially, the Permission service was created to handle all the logic, but we found clever way to rely on JwtClaimsTransformer and offload to FastEndpoints Permissions config.

I have added a note to the `RBAC-TODOs.md` to clean up `IPermissionService` public interface and API once we are closer to production. At this point we will really know what we will need and what nto

PR #492, opened 2025-10-06
feat: Endpoints for managing the roles of a user

dspetrov:
# Endpoints for managing the roles of a user

## Description
The three endpoints have been added:
- POST `/api/auth/users/{userId}/roles` – Assign role to user - the role name is in the request JSON body
- DELETE `/api/auth/users/{userId}/roles/{roleName}` – Remove role from user  
- GET `/api/auth/users/{userId}/roles` – Retrieve roles assigned to a user

The implementation includes the endpoints, the commands and queries the use, and `RoleManagementService`.

## Related Issues
closes https://github.com/endatix/endatix-private/issues/296

## Type of Change

## Checklist


PeterMarinov (approved):
Approved

>[!NOTE]
>**One note:**
>I am a bit concerned that more and more business logic is moved away from the handlers and into the services, which means the Infrastructure layer.
>
>I understand that this is due to the fact that the Core project is not aware of ASP.NET Core Identity, so we have two options:
>1. Leave it as it is and move business logic to services and other Clean Architecture glue
>2. Move User Management into a separate module similar to Agent module and avoid the need to have that many abstractions for simple CRUD + business logic
>
> ✅ I vote for option 2

PR #693, opened 2026-04-16
fix: "get form by id" endpoint doesn't return submissions count

orilux:
## Description
Fixes GET form/{id} was not returning the submissions count

Addresses https://github.com/endatix/endatix-hub/issues/559

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

### DIFF
--- a/src/Endatix.Core/Specifications/FormByIdWithSubmissionsCountSpec.cs
+++ b/src/Endatix.Core/Specifications/FormByIdWithSubmissionsCountSpec.cs
@@ -1,0 +1,31 @@
+using Ardalis.Specification;
+using Endatix.Core.Entities;
+using Endatix.Core.UseCases.Forms;
+
+namespace Endatix.Core.Specifications;
+
+public sealed class FormByIdWithSubmissionsCountSpec : Specification<Form, FormDto>
+{
+    public FormByIdWithSubmissionsCountSpec(long formId)
+    {
+        Query
+            .Where(form => form.Id == formId)
+            .AsNoTracking();
+
+        Query.Select(form =>
+            new FormDto()
+            {
+                Id = form.Id.ToString(),
+                Name = form.Name,
+                Description = form.Description,
+                IsEnabled = form.IsEnabled,
+                IsPublic = form.IsPublic,
+                ThemeId = form.ThemeId.HasValue ? form.ThemeId.Value.ToString() : null,
+                ActiveDefinitionId = form.ActiveDefinitionId.HasValue ? form.ActiveDefinitionId.Value.ToString() : null,
+                CreatedAt = form.CreatedAt,
+                ModifiedAt = form.ModifiedAt,
+                SubmissionsCount = form.FormDefinitions.SelectMany(fd => fd.Submissions).Count(),
+                WebHookSettingsJson = form.WebHookSettingsJson
+            });
+    }
+}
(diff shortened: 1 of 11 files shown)


src/Endatix.Core/Specifications/FormByIdWithSubmissionsCountSpec.cs
```diff
@@ -0,0 +1,30 @@
...
+
+public sealed class FormByIdWithSubmissionsCountSpec : Specification<Form, FormDto>
+{
+    public FormByIdWithSubmissionsCountSpec(long formId)
+    {
+        Query
+            .Where(form => form.Id == formId);
+
+        Query.Select(form =>
+            new FormDto()
```
PeterMarinov: This logic is almost the same as the one in FormsWithSubmissionsCountSpec with the only difference that there, we count all submissions
 
`SubmissionsCount = form.FormDefinitions.SelectMany(fd => fd.Submissions).Count()` 
and here we count only non-deleted.
`SubmissionsCount = form.FormDefinitions.SelectMany(fd => fd.Submissions).Count(s => !s.IsDeleted),`

Since we do soft delete on submission, the new logic is correct and the old one is bug.

My suggestion:
- use mapper that centralizes this ToFormDto logic. You can add static class. I have in a PR not pushed yet

```cs
using System.Linq.Expressions;
using Endatix.Core.Entities;
using Endatix.Core.UseCases.Forms;

namespace Endatix.Core.Specifications;

/// <summary>
/// Reusable LINQ projections for mapping <see cref="Form"/> to DTOs.
/// Keep projections as expression trees so EF/specification queries can translate them to SQL.
/// </summary>
public static class FormProjections
{
    /// <summary>
    /// Maps a form to <see cref="FormDto"/> and includes total submissions count.
    /// </summary>
    public static readonly Expression<Func<Form, FormDto>> ToFormDtoWithSubmissionsCount = form => new FormDto
    {
        Id = form.Id.ToString(),
        Name = form.Name,
        Description = form.Description,
        IsEnabled = form.IsEnabled,
        IsPublic = form.IsPublic,
        ThemeId = form.ThemeId.HasValue ? form.ThemeId.Value.ToString() : null,
         ActiveDefinitionId = form.ActiveDefinitionId.HasValue ? form.ActiveDefinitionId.Value.ToString() : null,
         CreatedAt = form.CreatedAt,
         ModifiedAt = form.ModifiedAt,
         SubmissionsCount = form.FormDefinitions.SelectMany(fd => fd.Submissions).Count(s => !s.IsDeleted),
         WebHookSettingsJson = form.WebHookSettingsJson
    };
}

// then in your Specification and the FormsWithSubmissionsCountSpec just do
Query.Select(FormProjections.ToFormDtoWithSubmissionsCount);
```

PeterMarinov (changes requested):
Please do the suggested change before we merge to clean up the Dto code. I hope it will be very quick

src/Endatix.Core/Specifications/FormByIdWithSubmissionsCountSpec.cs
orilux: @PeterMarinov I just realized that `s => !s.IsDeleted` is redundant, because we have the [Global Query Filter here](https://github.com/endatix/endatix/blob/3bffd3d978d591d81cb6b981e98c8b293c73e841/src/Endatix.Infrastructure/Data/DbContextModelBuilderExtensions.cs#L35). We should probably remove the additional ones altogether. Any thoughts?

PeterMarinov: sure, I have forgotten about the filter. 

Another reason to extract the common DTO logic to single projection file

PeterMarinov (approved):
Approved

PR #716, opened 2026-04-30
feat: Single submission administration

orilux:
## Description
- Added optional single-submission gate support on forms via `LimitOnePerUser`.
- Added API/request/DTO/mapping support for form metadata, single-submission state, and test-submission flags.
- Enforced gate rules:
  - cannot enable on public forms
  - cannot disable once enabled
  - cannot enable when existing eligible duplicate submissions exist
- Preserved existing duplicate submissions without marking them as test data.
- Removed the global unique submission index and replaced it with transactional/race-safe create handling.
- Added serializable transaction handling for gated submissions, including conflict handling for duplicate races.
- Added `isTestSubmission` support for submission creation, listing, filtering, and API responses.
- Added form test permission migration.
- Added/updated tests for form update rules, partial update rules, active definition gating, submission creation races, and submission filtering.

## Related Issues
https://github.com/endatix/endatix-hub/issues/580

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Release Notes

* **New Features**
  * Added ability to filter submissions by test status in submission lists.

* **Bug Fixes**
  * Improved error handling for form updates with clearer conflict responses.
  * Added validation for single-submission-per-user gate to prevent invalid state transitions (e.g., disabling an enabled gate, making single-submission forms public, enabling when duplicates exist).


src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs
```diff
@@ -109,14 +102,72 @@ public async Task<Result<Submission>> Handle(CreateSubmissionCommand request, Ca
...
-        try
+        var shouldEnforceSingleSubmissionGate =
+            formWithActiveDefinition.LimitOnePerUser &&
+            !string.IsNullOrWhiteSpace(request.SubmittedBy) &&
+            !canBypassSingleSubmissionLimit;
+
+        if (shouldEnforceSingleSubmissionGate)
         {
-            await submissionRepository.AddAsync(submission, cancellationToken);
+            await unitOfWork.BeginTransactionAsync(cancellationToken, IsolationLevel.Serializable);
```
PeterMarinov: Solution is great to ensure data integrity, but I have concerns with the concurrency and perofrmance in high traffic bursts.

The current approach utilizing `IsolationLevel.Serializable` to prevent duplicate form submissions introduces critical scalability and performance bottlenecks. Elevating the transaction isolation level to resolve application-level race conditions (Phantom Reads) is not recommended for high-throughput APIs.

### Summary
Potential Issues with **Serializable** transactions: This level can lead to significant blocking and potential deadlocks, reducing concurrency and application performance - ([link](https://medium.com/@serhatalftkn/understanding-transaction-isolation-levels-in-entity-framework-core-89d8e89f0ec4))

### The Case Against Serializable Transactions
1. **Lock Escalation & Concurrency Drop:** `Serializable` transactions acquire range locks on indexes to prevent phantom reads. Under concurrent load, SQL engines (Postgres/SQL Server) often escalate these to table-level locks. This effectively serializes *all* inserts to the `Submissions` table, severely degrading API throughput.
2. **Deadlock Probability:** If two concurrent transactions execute the `AnyAsync()` check before either attempts an `Insert`, they will both hold read locks that block the other's write lock, resulting in guaranteed deadlocks and forcing the application to implement arbitrary retry loops.
3. **Resource Starvation:** Transactions held open while waiting for locks consume connection pool threads, potentially leading to widespread API timeouts under spike traffic.

### Recommended Approach: Domain-Driven Projected Unique Index
Push the concurrency control down to the storage engine's B-Tree using a Semantic Unique Key. This provides an O(log N) insert with absolute data integrity and zero explicit application locks (allowing the default `ReadCommitted` isolation level).

Indexes offer exactly what we need for this hottest write path. They use B-trees, which are
> Short-term share/exclusive page-level locks are used for read/write access. Locks are released immediately after each index row is fetched or inserted. These index types provide the highest concurrency without deadlock conditions.
[Info here](https://www.postgresql.org/docs/current/locking-indexes.html)

#### Implementation Directions

Please refactor the implementation to distribute the logic across the Clean Architecture boundaries as follows:

**1. Domain Layer ([[src/Endatix.Core/Entities/Submission.cs](https://www.google.com/search?q=https://github.com/endatix/endatix/blob/main/src/Endatix.Core/Entities/Submission.cs)](https://github.com/endatix/endatix/blob/main/src/Endatix.Core/Entities/Submission.cs))**
The Domain dictates the shape of the restriction. Introduce a nullable string property `RestrictionKey`.
* Calculate and populate this key in the constructor *only* if the form enforces the one-per-user restriction, the submission is not a test, and the user is authenticated.
* The key should be deterministic (e.g., `$"UserLimit_Form{formId}_User{submittedBy}"`). Leave it `null` if the rule does not apply.

**2. Infrastructure Layer ([[src/Endatix.Infrastructure/Data/Config/AppEntities/SubmissionConfiguration.cs](https://www.google.com/search?q=https://github.com/endatix/endatix/blob/main/src/Endatix.Infrastructure/Data/Config/AppEntities/SubmissionConfiguration.cs)](https://github.com/endatix/endatix/blob/main/src/Endatix.Infrastructure/Data/Config/AppEntities/SubmissionConfiguration.cs))**
Enforce the integrity at the database schema level.
* Configure a `HasIndex` on the `RestrictionKey`.
* Apply `.IsUnique()`.
* Apply `.HasFilter("[RestrictionKey] IS NOT NULL")` (or equivalent Postgres syntax if using Npgsql) to ensure forms without this restriction can accept infinite `null` values.

**3. UseCase Layer ([[src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs](https://www.google.com/search?q=https://github.com/endatix/endatix/blob/main/src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs)](https://github.com/endatix/endatix/blob/main/src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs))**
The Handler orchestrates the domain and gracefully handles the database fallback.
* **Remove** the explicit `IsolationLevel.Serializable` transaction block entirely.
* **Keep** the optimistic `AnyAsync()` check (the "soft" check) to reject duplicate submissions without throwing exceptions during normal, non-concurrent operations.
* **Add** a `try/catch` block around `_unitOfWork.SaveChangesAsync(cancellationToken)`. Catch the `DbUpdateException` (the "hard" check) which gets thrown if two requests bypass the soft check simultaneously. Return a `Result.Conflict("A submission already exists for this user.")`.


### Additional resources

- [Understanding Transaction Isolation Levels in Entity Framework Core](https://atalupadhyay.wordpress.com/2025/01/20/postgresql-serializable-isolation-with-net-a-deep-dive-with-hands-on-lab/)

src/Endatix.Api/Endpoints/Forms/Update.cs
```diff
@@ -30,7 +32,7 @@ public override void Configure()
     }
 
     /// <inheritdoc/>
-    public override async Task<Results<Ok<UpdateFormResponse>, BadRequest, NotFound>> ExecuteAsync(UpdateFormRequest request, CancellationToken cancellationToken)
+    public override async Task<Results<Ok<UpdateFormResponse>, BadRequest, NotFound, ProblemHttpResult>> ExecuteAsync(UpdateFormRequest request, CancellationToken cancellationToken)
```
PeterMarinov: You don't need BadRequest, NotFound here. Just leave `Task<Results<Ok<UpdateFormResponse>, ProblemHttpResult>> ` and problem results handles all other cases

src/Endatix.Api/Endpoints/Forms/PartialUpdate.cs
```diff
@@ -10,7 +12,7 @@ namespace Endatix.Api.Endpoints.Forms;
 /// <summary>
 /// Endpoint for partially updating a form.
 /// </summary>
-public class PartialUpdate(IMediator mediator) : Endpoint<PartialUpdateFormRequest, Results<Ok<PartialUpdateFormResponse>, BadRequest, NotFound>>
+public class PartialUpdate(IMediator mediator) : Endpoint<PartialUpdateFormRequest, Results<Ok<PartialUpdateFormResponse>, BadRequest, NotFound, ProblemHttpResult>>
```
PeterMarinov: Same here  => we only need 
`Endpoint<PartialUpdateFormRequest, Results<Ok<PartialUpdateFormResponse>, ProblemHttpResult>>`

PeterMarinov (changes requested):
Review completed. It's great to see this taking shape. 

My main concern is with the write path of the Submission create where we should pay extra care as this is our most critical path. 

Please consider the comments, resources and suggestions. I could not find real data and evidence besides recommendations of vendors or experts - no benchmarks, but I think it's work review and plan in this or another PR.

src/Endatix.Api/Endpoints/Forms/Update.cs
orilux: Done

src/Endatix.Api/Endpoints/Forms/PartialUpdate.cs
orilux: Done

src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs
orilux: Agreed and decided to address immediately to avoid the complications of making such an architectural change after the feature has hit production.

What changed:
- Added nullable `Submission.RestrictionKey` and deterministic key creation in [Submission.cs](C:/Code/endatix-saas/oss/src/Endatix.Core/Entities/Submission.cs).
- Removed the `Serializable` transaction path from [CreateSubmissionHandler.cs](C:/Code/endatix-saas/oss/src/Endatix.Core/UseCases/Submissions/Create/CreateSubmissionHandler.cs).
- Kept the soft `AnyAsync` duplicate check, then relies on the filtered unique `RestrictionKey` index as the hard concurrency guard.
- Added provider-specific unique filtered indexes for PostgreSQL and SQL Server.
- Updated migrations/snapshots so existing submissions get `RestrictionKey = null`; no existing duplicates are marked as test, and the migration should not fail on old duplicate submitters.
- Reverted the `IUnitOfWork.BeginTransactionAsync` isolation-level overload since it is no longer needed.
- Added/updated tests for handler behavior and `Submission` restriction-key creation.

Small deviations from the recommended approach:
- I did not make `Submission` independently inspect the `Form` policy. The handler computes whether the gate applies and passes `EnforceSingleSubmissionGate` through `SubmissionCreateOptions`; the entity still owns the key shape. This avoids coupling `Submission` construction to the full `Form` aggregate.
- I catch `DuplicateSubmissionException` instead of raw `DbUpdateException` in the handler. `AppDbContext` already translates provider-specific unique violations for `UX_Submissions_RestrictionKey` into that domain exception, and `repository.AddAsync` is the save boundary here.
- I put the filtered index in provider-specific config because PostgreSQL and SQL Server filter syntax differs.

PeterMarinov (approved):
Great. Ready to go

PR #915, opened 2026-07-28
fix(nuget): decouple Endatix.Hosting from the Reporting module

igdragoev:
## Goal

`Endatix.Hosting` cannot be restored from nuget.org by any external consumer. `dotnet add package Endatix.Hosting` fails outright:

```
error NU1101: Unable to find package Endatix.Modules.Reporting.
No packages exist with this id in source(s): nuget.org
```

## Cause

`Endatix.Hosting` took a `ProjectReference` on `Endatix.Modules.Reporting`, which is `IsPackable=false`. **NuGet turns a `ProjectReference` into a package dependency regardless of `IsPackable`**, so the published nuspec declared a dependency on an id that was never pushed — it 404s.

This is invisible inside the repo, because `ProjectReference`s always resolve locally. It only bites consumers, after the release is already public.

## Fix

Rather than publish the implementation, invert the direction so the hosting library no longer knows about the module.

Modules already declare optional capabilities through `IHasFeatureFlag` and `IHasDbMigrations`. This adds **`IHasFastEndpointsConfig`** in the same spirit and lets `UseModule` apply it:

```csharp
if (module is IHasFastEndpointsConfig endpointConfigModule)
{
    Api.ConfigureFastEndpoints(endpointConfigModule.ConfigureFastEndpoints);
}
```

`ReportingModule` implements it, so the host registers the module exactly the way any other module is registered, with no module-specific code left in `Endatix.Hosting`:

```csharp
builder.Host.ConfigureEndatixWithDefaults(endatix => endatix.UseModule(ReportingModule.Instance));
```

**The capability lives in `Endatix.Api`, not `Endatix.Framework`** — `Framework` has no FastEndpoints dependency and shouldn't gain one, while any module exposing endpoints already references `Api`. This also means a module never has to reference `Endatix.Hosting`, which matters beyond layering: `Endatix.Hosting` pulls in **both** persistence providers, so a module referencing it would drag SQL Server and PostgreSQL into every consumer's dependency closure.

Also drops a redundant guard — the previous code called `ShouldRegister` a second time around the FastEndpoints wiring, but `UseModule` already evaluates it and returns early when the flag is off.

## Why Contracts is published but the implementation is not

`Endatix.Modules.Reporting.Contracts` is the **extension point**, not the implementation: `Endatix.Api`'s submission export endpoint depends on it and degrades gracefully when nothing is registered. Publishing it is what lets *any* implementation plug in. It contains 30 files of interfaces, DTOs and settings types.

## Regression guard

`release-prepare.sh` now fails the build if a packed package declares a dependency on a repo-owned project that was not packed. Only repo-owned ids are checked, so packages like `Endatix.Outbox.Engine` that ship from their own repo are correctly ignored.

## Verification

- `dotnet build -c Release` — **0 errors**
- `dotnet pack` — 9 packages; `Contracts` **included**, implementation **excluded**
- `Endatix.Hosting` nuspec now lists only `Api`, `Framework`, `Infrastructure`, `Persistence.PostgreSql`, `Persistence.SqlServer` — no Reporting
- **A scratch project referencing `Endatix.Hosting` from a local feed restores successfully.** This is the regression proof: before this change the same project fails with `NU1101`
- The new guard passes against the packed feed
- Tests: Hosting 19/19 · Reporting 382/383 (1 pre-existing skip) · Api 637/637 · Core 1022/1022

One pre-existing failure in `Endatix.Infrastructure.Tests` (`DefaultCsvFormatterTests`, a decimal-separator/locale assertion) reproduces identically on clean `main` and is unrelated.

## No runtime change

Reporting stays gated behind `Endatix:FeatureFlags:ReportingModule`, off when the key is absent, and the OSS image still registers it exactly as before. `ReportingModuleStartupMigrationTests` boots `WebApplicationFactory<EndatixWebHost::Program>`, which is where registration now happens.

### DIFF
--- a/src/Endatix.WebHost/Program.cs
+++ b/src/Endatix.WebHost/Program.cs
@@ -1,8 +1,11 @@
 using Endatix.Hosting;
+using Endatix.Modules.Reporting;
 
 var builder = WebApplication.CreateBuilder(args);
 
-builder.Host.ConfigureEndatix();
+// Modules are registered by the host, not by Endatix.Hosting, so the hosting library stays
+// independent of them. UseModule honours Endatix:FeatureFlags:ReportingModule.
+builder.Host.ConfigureEndatixWithDefaults(endatix => endatix.UseModule(ReportingModule.Instance));
 
 var app = builder.Build();
 
--- a/scripts/release-prepare.sh
+++ b/scripts/release-prepare.sh
@@ -23,6 +23,27 @@ dotnet build -c Release --no-restore -p:Version="${VERSION}"
 echo "──── Packing NuGet packages at version ${VERSION} ────"
 dotnet pack -c Release --no-build -p:Version="${VERSION}" -o build/packages/nuget
 
+# A published package must never depend on one of ours that was not packed with
+# it. NuGet turns a ProjectReference into a package dependency even when the
+# referenced project sets IsPackable=false, so the nuspec ends up pointing at an
+# id nobody ever pushed and every external `dotnet add package` dies with NU1101
+# — invisible in this repo (ProjectReferences always resolve) until the release
+# is already public. Only repo-owned ids are checked: a dependency such as
+# Endatix.Outbox.Engine ships from its own repo and is legitimately absent here.
+echo "──── Verifying packed dependencies are self-contained ────"
+UNPACKED_DEPS=0
+for nupkg in build/packages/nuget/*."${VERSION}".nupkg; do
+  pkg_id="$(basename "${nupkg}" ".${VERSION}.nupkg")"
+  while read -r dep; do
+    [[ -n "${dep}" ]] || continue
+    [[ -f "src/${dep}/${dep}.csproj" ]] || continue
+    [[ -f "build/packages/nuget/${dep}.${VERSION}.nupkg" ]] && continue
+    echo "::error::${pkg_id} declares a dependency on ${dep}, which was not packed. Make src/${dep}/${dep}.csproj packable or drop the ProjectReference." >&2
+    UNPACKED_DEPS=1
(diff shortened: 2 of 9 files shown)


src/Endatix.WebHost/Program.cs
```diff
@@ -1,8 +1,11 @@
 using Endatix.Hosting;
+using Endatix.Modules.Reporting;
 
 var builder = WebApplication.CreateBuilder(args);
 
-builder.Host.ConfigureEndatix();
+// Modules are registered by the host, not by Endatix.Hosting, so the hosting library stays
+// independent of them. UseModule honours Endatix:FeatureFlags:ReportingModule.
+builder.Host.ConfigureEndatixWithDefaults(endatix => endatix.UseModule(ReportingModule.Instance));
```
PeterMarinov, line 8: This means that all users not need to explicitly enable it to use it. Maybe we need to discuss this change before merging.

2 use cases that are of interest:
1. Existing customers, need to change their Program.cs, so we need to communicate this before merging
2. Enable with Docker for evaluation - if we can only do this with Env variable will make it easier or we have to do templates e.g. HostWithReporting vs StandardHost

PeterMarinov (changes requested):
@igdragoev - one comment regarding the braking change of the module on y default vs modification of the `Program.cs`

src/Endatix.WebHost/Program.cs
igdragoev: I think for now we will just publish the package, which is not breaking change. I will close this PR and will open a new one.

igdragoev:
As I said we decided to take a different approach. Reporting will be just published to fix dotnet restore step

PR #918, opened 2026-07-28
fix(nuget): publish the Reporting module packages

igdragoev:
## The problem

`Endatix.Hosting 0.7.0` **cannot be restored by any external consumer**. Not "a feature is broken" — the package is entirely uninstallable:

```
error NU1101: Unable to find package Endatix.Modules.Reporting.
No packages exist with this id in source(s): nuget.org
```

Only `0.7.0` is affected; `0.6.0` and earlier predate the Reporting module.

## Cause

`src/Endatix.Hosting` has a `ProjectReference` to `src/Endatix.Modules.Reporting`, and `src/Endatix.Api` one to `src/Endatix.Modules.Reporting.Contracts`. Both referenced projects set `IsPackable=false`.

**NuGet turns a `ProjectReference` into a package dependency regardless of `IsPackable`.** The published nuspecs therefore declare a dependency on ids that were never pushed. This is invisible inside the repo — `ProjectReference`s always resolve locally and in CI — and only surfaces once the release is public.

## The fix

Make both Reporting projects packable, with package metadata consistent with the rest of the repo, and give each the `README.md` that `Directory.Build.props` requires via `<PackageReadmeFile>` (packing without one fails `NU5019`).

### Auto-registration is deliberately preserved

The Reporting module **stays in this repo and ships in every install**. It is gated at **runtime** by `Endatix:FeatureFlags:ReportingModule`, not by being absent from the build.

So `EndatixBuilder.UseDefaults()` keeps its `UseModule(ReportingModule.Instance)` call, and `Endatix.Hosting.csproj` keeps its `ProjectReference` — legitimate now that the package is published. `Endatix.WebHost/Program.cs` and its csproj are untouched.

**No consumer has to change `Program.cs`. There is no breaking change.**

### `IHasFastEndpoints`: modules own their endpoints

`UseDefaults()` previously special-cased Reporting's endpoint wiring:

```csharp
UseModule(ReportingModule.Instance);
if (EndatixModuleRegistration.ShouldRegister(Configuration, ReportingModule.Instance))
{
    Api.ConfigureFastEndpoints(ReportingModule.ConfigureFastEndpoints);
}
```

`UseModule` already calls `ShouldRegister` and returns early when the flag is off, so the explicit check was redundant. It reduces to `UseModule(ReportingModule.Instance);` once the endpoint wiring moves onto the generic module path, behind a new capability interface:

```csharp
public interface IHasFastEndpoints
{
    // Defaults to the assembly declaring the module.
    IEnumerable<Assembly> EndpointAssemblies => [GetType().Assembly];

    // No-op by default.
    void ConfigureFastEndpoints(Config config) { }
}
```

The interface covers **endpoint discovery**, not just endpoint configuration. Previously `UseModule` called `Api.ScanAssemblies(module.Assembly)` for every module, whether or not it had any endpoints; now discovery follows the interface. Three consequences:

- A module can ship endpoints in a **satellite assembly** by overriding `EndpointAssemblies`.
- A module with endpoints but no serializer/OpenAPI needs implements **nothing** — both members are defaulted.
- A module with **no** endpoints no longer drags its assembly into the FastEndpoints scan list, nor adds another set of FastEndpoints singleton descriptors. (`AddFastEndpoints` uses `AddSingleton`, not `TryAdd`, so every `ScanAssemblies` call leaves another `EndpointData`/`Cfg`/`CommandHandlerRegistry` descriptor behind; correctness rests on last-registration-wins.)

The trade-off is that a module shipping endpoints *without* the interface would silently serve none of them. To keep that failure visible, `UseModule` logs a warning when a registered module's assembly declares concrete `IEndpoint` types but the interface is absent. That reflection runs only for modules that skip the interface — the assembly is not scanned otherwise.

The interface lives in **`Endatix.Api`, not `Endatix.Framework`**, because `Endatix.Framework` has no FastEndpoints dependency and must not gain one. It mirrors the existing `IHasFeatureFlag` / `IHasDbMigrations` capabilities.

It is shaped this way now rather than in a follow-up because it becomes **public API of `Endatix.Api`** the moment these packages publish.

### Release guard

`scripts/release-prepare.sh` now fails the build when a packed package declares a dependency on a repo-owned project that was not packed alongside it, so this class of break cannot reach a release again. Only repo-owned ids are checked — a dependency such as `Endatix.Outbox.Engine` legitimately ships from its own repo.

## Verification

`dotnet build -c Release` → **0 errors**.

`dotnet pack -c Release -p:Version=0.7.99-local` → **10 packages**, now including both `Endatix.Modules.Reporting` and `Endatix.Modules.Reporting.Contracts`.

`Endatix.Hosting.0.7.99-local.nupkg` nuspec, dependency now resolvable in-feed:

```
<dependency id="Endatix.Api" version="0.7.99-local" />
<dependency id="Endatix.Framework" version="0.7.99-local" />
<dependency id="Endatix.Infrastructure" version="0.7.99-local" />
<dependency id="Endatix.Modules.Reporting" version="0.7.99-local" />   <-- resolves
<dependency id="Endatix.Persistence.PostgreSql" version="0.7.99-local" />
<dependency id="Endatix.Persistence.SqlServer" version="0.7.99-local" />
```

**Regression proof.** A scratch consumer project referencing `Endatix.Hosting 0.7.99-local`, against a `nuget.config` with the local feed + nuget.org:

```
Restored .../Consumer.csproj (in 10.27 sec).       # exit 0
```

The same consumer against the same feed with only the two Reporting nupkgs removed — i.e. exactly what `main` produces — reproduces the reported failure:

```
error NU1101: Unable to find package Endatix.Modules.Reporting.
error NU1101: Unable to find package Endatix.Modules.Reporting.Contracts.
```

**Release guard**, run against the packed feed: passes. Run against the same feed minus the two Reporting packages, it catches both instances of the original bug and exits 1:

```
::error::Endatix.Api declares a dependency on Endatix.Modules.Reporting.Contracts, which was not packed.
::error::Endatix.Hosting declares a dependency on Endatix.Modules.Reporting, which was not packed.
```

**Tests.**

| Project | Result |
|---|---|
| `Endatix.Hosting.Tests` | 25 passed, 0 failed |
| `Endatix.Modules.Reporting.Tests` | 387 passed, 0 failed, 1 skipped |
| `Endatix.Api.Tests` | 637 passed, 0 failed |
| `Endatix.Core.Tests` | 1022 passed, 0 failed |

Six tests were added to `EndatixBuilderUseModuleTests`, covering both halves of the capability:

- configuration applied when the module implements the interface and the flag is on;
- not applied when the flag is off;
- no configurator registered for a module without the capability;
- the module's own assembly scanned by default;
- a declared satellite assembly scanned instead of the module's own;
- **nothing** scanned for a module without the interface.

Verified non-vacuous: restoring the old unconditional `Api.ScanAssemblies(module.Assembly)` fails the last two, and deleting the capability branch fails the first.


PeterMarinov (dismissed):
Approved

PR #921, opened 2026-07-29
fix(featureflags): shut OpenFeature down once per process

igdragoev:
## The problem

`Endatix.IntegrationTests.ReportingModuleStartupMigrationTests` has been failing on unrelated PRs — roughly **two runs in three** — with a failure that has nothing to do with the test's assertion:

```
System.Threading.Channels.ChannelClosedException : The channel has been closed.
   at System.Threading.Channels.ChannelWriter`1.Complete(Exception error)
   at OpenFeature.EventExecutor.ShutdownAsync()
   at OpenFeature.Api.ShutdownAsync()
   at OpenFeature.Hosting.Internal.FeatureLifecycleManager.ShutdownAsync(CancellationToken)
   at OpenFeature.Hosting.HostedFeatureLifecycleService.StoppedAsync(CancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.StopAsync(CancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.DisposeAsync()
```

The assertion passed — a failed `Assert.True` raises `TrueException`, not this. The host blew up during **teardown**.

## Cause

`OpenFeature.Api` is a **process-wide singleton**. The hosted service that shuts it down is registered **per host**.

`EventExecutor.ShutdownAsync` (OpenFeature 2.14.0) is:

```csharp
public async Task ShutdownAsync()
{
    this.EventChannel.Writer.Complete();
    await this.EventChannel.Reader.Completion.ConfigureAwait(false);
}
```

No guard, and a completed `Channel` cannot be reopened. The first host to stop closes it; the next host's `StoppedAsync` calls `Complete()` on a closed channel and throws.

A single application host never notices — there is nothing to shut down twice. Anything that builds more than one host in a process does. The integration suite builds three `EndatixWebApplicationFactory` instances (the collection fixture, plus one each in `ReportingModuleStartupMigrationTests` and `HealthCheckTests`), so whichever stops second fails on teardown. Which one that is depends on ordering, which is why it looks like flake.

## The fix

Decorate OpenFeature's `IFeatureLifecycleManager` so shutdown reaches the SDK at most once per process:

```csharp
public ValueTask ShutdownAsync(CancellationToken cancellationToken = default) =>
    Interlocked.Exchange(ref _shutdownStarted, 1) == 0
        ? inner.ShutdownAsync(cancellationToken)
        : ValueTask.CompletedTask;
```

Three deliberate choices:

- **Decorate rather than replace.** `FeatureLifecycleManager.EnsureInitializedAsync` does real work — default provider, named providers, hooks, event handlers. Only `ShutdownAsync` needed changing, so the rest is delegated untouched, along with OpenFeature's own lifecycle timing and its configurable `FeatureStartState`/`FeatureStopState`.
- **Initialization is not guarded.** It is per host and must run for every one; only shutdown is one-way.
- **The guard is static** because the resource it protects is static. Shutting OpenFeature down is a process-wide, one-way operation, so the second and later calls have nothing left to do.

Registration hooks in after `AddOpenFeature` and is idempotent — a second `AddEndatixOpenFeature()` finds the decorator already installed and does nothing.

### Also: an obsolete call removed

`builder.AddHostedFeatureLifecycle()` has been obsolete since OpenFeature 2.14 and was emitting `CS0618`:

> Calling AddHostedFeatureLifecycle() is no longer necessary. OpenFeature will inject this automatically when you call AddOpenFeature().

Removed. **This is not the fix** — the lifecycle service is registered either way, and `AddHostedService` de-duplicates by implementation type, so the call was redundant rather than harmful.

## Verification

`dotnet build` → 0 errors, and the `CS0618` warning is gone.

`Endatix.Infrastructure.Tests` → **1117 passed, 1 failed**. The one failure is the pre-existing culture-dependent CSV export bug, unrelated to this change and fixed separately in #919. Six new tests:

| Test | Covers |
|---|---|
| `ShutdownAsync_FirstCall_ShutsDownTheInnerManager` | the first shutdown still reaches OpenFeature |
| `ShutdownAsync_CalledTwice_ReachesTheInnerManagerOnce` | the guard |
| `ShutdownAsync_SecondHostInTheSameProcess_DoesNotThrow` | the reported failure, with an inner manager that throws `ChannelClosedException` like the real one |
| `EnsureInitializedAsync_EveryCall_ReachesTheInnerManager` | initialization is *not* suppressed |
| `AddEndatixOpenFeature_DecoratesTheLifecycleManager` | wiring |
| `AddEndatixOpenFeature_CalledTwice_RegistersASingleDecorator` | idempotent registration |

Verified non-vacuous: dropping the `Interlocked` guard fails the two shutdown tests.

**On the end-to-end proof, being precise:** the integration suite passes locally on this branch (65/65) — but it also passes locally *without* this change, so the local run proves nothing on its own. The failure only reproduces on CI hardware. The real check is CI, and #918 has been rebased onto this branch to run it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application shutdown handling to prevent duplicate shutdown attempts when multiple hosts run in the same process.
  * Preserved feature initialization behavior across repeated calls.
  * Improved resilience when shutdown is triggered more than once.

* **Tests**
  * Added coverage for repeated shutdowns, multiple hosts, initialization, and feature lifecycle registration.

### DIFF
--- a/src/Endatix.Infrastructure/FeatureFlags/SingleShutdownFeatureLifecycleManager.cs
+++ b/src/Endatix.Infrastructure/FeatureFlags/SingleShutdownFeatureLifecycleManager.cs
@@ -1,0 +1,44 @@
+using OpenFeature.Hosting;
+
+namespace Endatix.Infrastructure.FeatureFlags;
+
+/// <summary>
+/// Wraps OpenFeature's lifecycle manager so that shutdown runs at most once per process.
+/// </summary>
+/// <remarks>
+/// <para>
+/// <c>OpenFeature.Api</c> is a process-wide singleton, but the hosted lifecycle service that shuts
+/// it down is registered per host. <c>Api.ShutdownAsync()</c> completes the SDK's event channel, and
+/// a completed channel cannot be reopened, so a second host stopping in the same process throws
+/// <c>ChannelClosedException</c> out of <c>IHostedLifecycleService.StoppedAsync</c>.
+/// </para>
+/// <para>
+/// A single application host never notices — there is nothing to shut down twice. Anything that
+/// builds more than one host in a process does: the integration test suite creates several
+/// <c>WebApplicationFactory</c> instances, and whichever stops second fails on teardown even though
+/// the test itself passed.
+/// </para>
+/// <para>
+/// The guard is static because the resource it protects is static. Shutting OpenFeature down is a
+/// process-wide, one-way operation, so the second and later calls have nothing left to do.
+/// </para>
+/// </remarks>
+internal sealed class SingleShutdownFeatureLifecycleManager(IFeatureLifecycleManager inner)
+    : IFeatureLifecycleManager
+{
+    private static int _shutdownStarted;
+
+    public ValueTask EnsureInitializedAsync(CancellationToken cancellationToken = default) =>
+        inner.EnsureInitializedAsync(cancellationToken);
+
+    public ValueTask ShutdownAsync(CancellationToken cancellationToken = default) =>
+        Interlocked.Exchange(ref _shutdownStarted, 1) == 0
+            ? inner.ShutdownAsync(cancellationToken)
+            : ValueTask.CompletedTask;
+
+    /// <summary>
+    /// Clears the process-wide guard. Exists so tests can exercise the first-shutdown path more
+    /// than once; nothing in the application resets it.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


PeterMarinov (approved):
Approved

Issue #476, opened 2025-08-28
[FEATURE] Add OpenTelemetry Support for Distributed Tracing and Metrics


PeterMarinov:
## Objective
Extend the existing Endatix.Hosting package to include OpenTelemetry support for distributed tracing, metrics collection, and integration with Azure Monitor/Application Insights while maintaining the current Serilog-based structured logging.

## Background
- Endatix already has Azure Application Insights integration via Serilog sink in `EndatixLoggingBuilder`
- The project uses a builder pattern approach via `Endatix.Hosting` package for configuration
- Need to extend the existing telemetry capabilities with OpenTelemetry for better observability

### 1. Core OpenTelemetry Integration
- Add OpenTelemetry tracing and metrics support
- Implement ASP.NET Core, HTTP Client, Runtime, and Process instrumentation
- Configure resource attributes including service name, version, environment, and cloud region
- Export telemetry data to Azure Monitor/Application Insights

### 2. Configuration Strategy
**Default Behavior**: Enable basic tracing and metrics with sensible defaults

**Configurable Options**:
- Service name and version
- Cloud region and environment detection
- Custom resource attributes
- Azure Monitor connection string
- Instrumentation selection (ability to enable/disable specific instrumentations)
- Sampling configuration

### 3. Integration with Existing Infrastructure
- Extend `EndatixLoggingBuilder` or create new `EndatixTelemetryBuilder`
- Maintain compatibility with existing Azure Application Insights Serilog sink
- Follow the established builder pattern used throughout Endatix.Hosting
- Ensure no breaking changes to existing logging configuration

### 4. Implementation Approach
- Create extension methods for the Endatix builder pattern
- Add new configuration section in appsettings.json for telemetry settings
- Implement both "UseDefaults" and granular configuration options
- Add appropriate NuGet package dependencies

## Reference POC Implementation
```csharp
builder.Services.AddOpenTelemetry()
    .ConfigureResource(resource => resource
       .AddService(serviceName: "ZeroLineMonitor", serviceVersion: "1.0.0")
       .AddAttributes(new Dictionary<string, object>
       {
           ["deployment.environment"] = builder.Environment.EnvironmentName,
           ["cloud.region"] = "North Europe"
       }))

    .WithTracing(tracing => tracing
        .AddAspNetCoreInstrumentation()
        .AddHttpClientInstrumentation())

    .WithMetrics(metrics => metrics
        .AddAspNetCoreInstrumentation()
        .AddHttpClientInstrumentation()
[...]

PR #944, opened 2026-08-06
fix(e476): repair OTLP startup crash + make the local telemetry dashboard reachable (PR-T4)

igdragoev:
# Make the local telemetry dashboard reachable (e476 PR-T4, compose half)

## Description

The [e476 plan](https://github.com/endatix/endatix-saas/blob/main/.cursor/plans/e476-opentelemetry-plan.md) calls for adding an Aspire Dashboard so traces, metrics and logs are visible locally without standing up Loki, Tempo and Grafana.

**It is already here.** `docker-env-dashboard` exists, and `endatix-api` and `endatix-hub` both already export OTLP to it on `:18889`. So the plan's premise was wrong, and the real reason nobody uses it is narrower and more mundane:

**You cannot open it.** Port `18888` was only `expose`d — which publishes to other containers on the network, not to the host — so there has never been a URL to visit. The dashboard is also mentioned **nowhere** in `docker/README.md`, `setup.sh` or `setup.bat` (zero matches), so nothing told you it existed. It has been quietly receiving telemetry that no one could look at.

### What changed

**1. Publish `18888` and `18889`, bound to `127.0.0.1`.**

The loopback bind is not incidental. The dashboard renders every request, log line and form payload the platform handles, and this compose file is what the README instructs people to `curl` and run — a bare `"18888:18888"` would put an unauthenticated telemetry UI on their LAN. `18889` is published too so an API or Hub run *outside* Docker (`dotnet run`, an IDE) can export to the same dashboard.

**2. `DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS: "true"`.**

Without it the dashboard mints a token and prints a login URL into its own logs, so step one of using it is `docker logs`. This flag is only defensible because of the loopback bind above, and the comment in the file says exactly that.

**3. Move off the nightly image.** `mcr.microsoft.com/dotnet/nightly/aspire-dashboard` publishes unreleased builds daily. Now the stable repo, pinned to `9.5.2` — matching how `postgres:17.6` is pinned in the same file.

**4. Document it.** The README listed only 8080 and 3000 under required ports and never named the dashboard. It now covers both new ports, the URL, what the UI is for, why it must stay on localhost, and how to point a locally-run service at it.

## Related Issues

Refs #476

## Type of Change

- [x] Documentation update
- [x] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Verification

Exercised against the real `9.5.2` image rather than reasoned about:

| Check | Result |
|---|---|
| `compose config` parses | ✅ both mappings render as `127.0.0.1:18888:18888` / `127.0.0.1:18889:18889` |
| UI reachable | ✅ `/` → `302` → `/structuredlogs` → `200`, 4401 bytes |
| No login required | ✅ no login page, no token line in the container logs |
| Flag is load-bearing | ✅ **without** it the logs carry `Login to the dashboard at .../login?t=953a0627…` |
| Not LAN-exposed | ✅ `podman port` reports `18888/tcp -> 127.0.0.1:18888` only |

## Notes for the reviewer

- **The image is pinned, so it will go stale.** I followed the `postgres:17.6` precedent in this file rather than tracking `latest`, on the grounds that a quickstart people download should be reproducible. Easy to change if you'd rather it float.
- **This is the compose half of PR-T4 only.** Hub → API `traceparent` propagation (AC12) is the other half and stays separate: `OtelTelemetryStrategy` registers `instrumentation-fetch`, which is *browser* instrumentation and does little inside a `NodeSDK`, while `HttpInstrumentation` patches `node:http` that undici-based `fetch` bypasses. Whether Next's own fetch tracing covers it is genuinely unknown and needs verifying against a running dashboard — which this PR is the prerequisite for.
- **It also closes an evidence gap in #940.** AC1 (`process_runtime_dotnet_*` and `http_server_*` reaching a collector) and AC5 (probes producing no spans) cannot be asserted by unit tests. With a reachable dashboard they become observable, and the `MetricReader` integration test the plan lists finally has somewhere to point.

---

# Added: fix a startup crash on `main` (`d09cb03`)

While verifying the dashboard against merged `main`, the API failed to start:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'value')
   at OpenTelemetry.Exporter.OtlpExporterOptions.set_Endpoint(Uri value)
   at EndatixTelemetryBuilder.ConfigureOtlp(...)
   at EndatixTelemetryBuilder.<Build>b__3(OtlpExporterOptions, MetricReaderOptions)
```

**Any host that sets `OTEL_EXPORTER_OTLP_ENDPOINT` crashes at startup on current `main`.** That is the entire intended production path — the Helm chart, this compose file, and the deployed API all configure telemetry exactly that way — so #940 shipped telemetry that cannot be switched on.

**Cause.** The review fix on #940 made `Build()` pass a *null* endpoint when the value came from an `OTEL_*` variable — deliberately, so the SDK reads the variable itself and applies the full specification (signal-specific overrides, per-signal path suffixes). But `ConfigureOtlp` still declared a non-nullable `Uri` and assigned it unconditionally, and `OtlpExporterOptions.Endpoint` rejects null. The parameter is now `Uri?` with a guarded assignment, which is what the null always meant: leave it alone.

**Why it reached `main`** — both worth fixing in how tests are written, not just in the code:

1. **The `AddOtlpExporter(...)` callback is deferred** and never ran in any test, because every existing test stopped at registration without building a provider. The same deferred-callback trap #940 documented for *validation*, hit from the other side. The five new tests build the `ServiceProvider` and resolve `MeterProvider`/`TracerProvider` so the callback executes — covering the global env var, both signal-specific ones, a configuration-sourced endpoint, and all instrumentation disabled.
2. **The compiler had already caught it.** Passing `Uri?` to a `Uri` parameter is `CS8604` and it was in the build output — but the command used to check that build piped through `grep -v "warning CS"`, filtering away the one warning that predicted this exact crash.

**Verified as a real guard**, not just a passing test: with the source fix reverted, **4 of the 5 new tests fail with the production `ArgumentNullException`**. The one that passes is the configuration-sourced endpoint, correctly — that path assigns a real `Uri`.

67 tests pass, up from 62.

> [!IMPORTANT]
> This PR now spans two concerns: the compose/README change and a source fix for a crash on `main`. They were kept together because the crash was found *by* the dashboard work and blocks the verification it exists to enable — but say the word and I'll split the fix into its own PR for a faster merge.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -3,7 +3,7 @@ services:
   # Telemetry UI for the OTLP data endatix-api and endatix-hub export below: traces,
   # metrics and structured logs in one container, with no Loki/Tempo/Grafana to stand up.
   docker-env-dashboard:
-    image: "mcr.microsoft.com/dotnet/aspire-dashboard:9.5.2"
+    image: "mcr.microsoft.com/dotnet/aspire-dashboard:13.4.2"
     container_name: "endatix-dashboard"
     environment:
       # Dev-only. Without it the dashboard mints a login token and prints it to its own
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -1,10 +1,22 @@
 name: "endatix-app"
 services:
+  # Telemetry UI for the OTLP data endatix-api and endatix-hub export below: traces,
+  # metrics and structured logs in one container, with no Loki/Tempo/Grafana to stand up.
   docker-env-dashboard:
-    image: "mcr.microsoft.com/dotnet/nightly/aspire-dashboard:latest"
-    expose:
-      - "18888"
-      - "18889"
+    image: "mcr.microsoft.com/dotnet/aspire-dashboard:9.5.2"
+    container_name: "endatix-dashboard"
+    environment:
+      # Dev-only. Without it the dashboard mints a login token and prints it to its own
+      # logs, so opening the UI means digging through `docker logs` first. Safe here ONLY
+      # because both ports below are bound to loopback -- never pair this with a 0.0.0.0
+      # publish, and never set it in a deployed manifest.
+      DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS: "true"
+    ports:
+      # Bound to 127.0.0.1 deliberately: the dashboard is unauthenticated (above) and
+      # renders every request, log line and form payload the platform handles. A bare
+      # "18888:18888" would publish that to the whole LAN.
+      - "127.0.0.1:18888:18888" # Dashboard UI
+      - "127.0.0.1:18889:18889" # OTLP/gRPC -- published so an API or Hub run outside Docker can export here too
     networks:
       - "endatix-net"
     restart: "always"
--- a/docker/README.md
+++ b/docker/README.md
@@ -57,6 +57,8 @@ curl -o docker-compose.yaml https://raw.githubusercontent.com/endatix/endatix/ma
 Make sure these ports are free on your machine:
 - **8080** - Endatix API  
 - **3000** - Endatix Hub
+- **18888** - Telemetry dashboard (localhost only)
+- **18889** - Telemetry ingest, OTLP/gRPC (localhost only)
 
 ### Step 3: Run the Setup Script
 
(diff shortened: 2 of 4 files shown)


docker/docker-compose.yaml
```diff
@@ -1,10 +1,22 @@
 name: "endatix-app"
 services:
+  # Telemetry UI for the OTLP data endatix-api and endatix-hub export below: traces,
+  # metrics and structured logs in one container, with no Loki/Tempo/Grafana to stand up.
   docker-env-dashboard:
-    image: "mcr.microsoft.com/dotnet/nightly/aspire-dashboard:latest"
-    expose:
-      - "18888"
-      - "18889"
+    image: "mcr.microsoft.com/dotnet/aspire-dashboard:9.5.2"
```
PeterMarinov: Latest tag is 13 - https://mcr.microsoft.com/artifact/mar/dotnet/aspire-dashboard/tags

The `DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS` is supported there too. I think `latest` tag is acceptable here or if not we can at set to 13

PeterMarinov (dismissed):
Approve. I think, before merging we must change the version

docker/docker-compose.yaml
igdragoev: Good catch — bumped to `13.4.2` in 5b93e78.

I verified the two things this block depends on against the actual image rather than the tag list, since a major bump is exactly where they'd break:

- **`DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS` still works** — the UI 302s straight to `/structuredlogs` rather than a login page, and there are zero token lines in the container logs. You were right.
- **Ports are unchanged** — `ASPNETCORE_URLS=http://+:18888` and `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL=http://+:18889`, so the two `OTEL_EXPORTER_OTLP_ENDPOINT` values further down still resolve.

One thing worth knowing: **13.x adds two more listeners** that 9.5.2 didn't have — OTLP/HTTP on 18890 and an MCP endpoint on 18891. I deliberately left both unpublished. They're only reachable inside `endatix-net`, and given the dashboard runs unauthenticated I'd rather not widen the surface without a reason to.

On `latest` vs `13` vs the exact patch — I went with `13.4.2` to match `postgres:17.6` right below it. This bump is itself the argument against a floating tag: a future major could move those ports or rename the auth switch, and with `latest` that lands on everyone's machine with no diff to point at. Happy to change it to `13` if you'd rather trade that for picking up patches automatically — it's a one-word change and I don't feel strongly.

PeterMarinov (approved):
Approved, thanks!

PR #950, opened 2026-08-10
docs(e476): observability and bring-your-own-telemetry pages (PR-T5)

igdragoev:
## Description

The documentation half of #476. Two new pages under **Configuration**, plus an architecture decision record.

Written on the assumption that **[#948](https://github.com/endatix/endatix/pull/948) (PR-T6, file logging) merges first** — see the ordering note below.

### `observability.mdx` — the operational page

The OTLP contract and everything around it: the env-var table including signal-specific variants, why a malformed endpoint fails startup rather than exporting nothing quietly, the provider-scoped level sections, file logging, the local Aspire dashboard, a self-hosted collector config, and a key-by-key **Serilog → Logging migration table** covering levels, sinks and the two removed code hooks.

Two things in there are traps rather than reference material, and are called out as such:

- **Levels default to `Warning`.** Exporting without lifting `Endatix` ships almost nothing, and the pipeline looks broken when it is merely quiet.
- **Relative file paths resolve against the content root.** The configuration this replaces used `/logs` at the filesystem root, so anyone copying their old absolute path forward should check it.

### `bring-your-own-telemetry.mdx` — what replaces shipping an exporter

Endatix ships no vendor exporters, so this page has to carry the weight that an `Azure.Monitor` dependency otherwise would. Worked examples for Azure Monitor and plain OTLP, `UseAzureMonitor()` **with no argument** (it resolves `APPLICATIONINSIGHTS_CONNECTION_STRING` itself — the page explicitly says not to invent an Endatix key), and why `builder.Logging.AddX()` in `Program.cs` silently does not survive.

It also carries the **cost material**, because this is the page where someone decides to switch App Insights on:

- Billing is **per GB ingested**; **dependency telemetry dominates** for a DB-heavy API, and every EF Core query becomes a billable record
- Enabling the distro is an **expansion** of what is collected, not a like-for-like swap — the old Serilog sink sent log records only
- **Log records belonging to unsampled traces are dropped by default**, so a naive `SamplingRatio = 0.1` discards ~90% of logs as well as traces — quite possibly including the error being diagnosed

## A pre-existing bug found while working

`infrastructure-configuration.mdx` documented Serilog **and an API that does not exist**:

```csharp
endatix.Infrastructure.ConfigureLogging(logging => {
    logging.UseConsole = true;  // no such thing
});
```

`ConfigureLogging` appears **nowhere in the codebase** — `grep` across `src/` returns nothing. That example never compiled, independently of anything in #476. Replaced with a short summary and a link to the new page.

## Also

- **`ARCHITECTURE.md`** — an Observability section (three signals one mechanism, env-vars authoritative, why the log exporter lives in the telemetry builder, why the builder ordering is load-bearing, Serilog's remaining role, no vendor exporters) and a decision-log entry.
- **`MigrationDocsUrl`** now points at the observability page. It was aimed at the configuration index as a deliberate placeholder because the page did not exist; it does now.

## Verification

**The Docusaurus site builds clean** — `onBrokenLinks: "throw"`, so the cross-page links and the `#migrating-from-the-serilog-section` anchor are verified by the build rather than by eye. No warnings.

84 tests still pass on the one-line code change.

> [!IMPORTANT]
> **Merge after [#948](https://github.com/endatix/endatix/pull/948).** These pages document `Endatix:Logging:File` and `Logging:EndatixFile:LogLevel`, which ship in that PR. Merging this first would publish documentation for a feature not in the release. No file conflicts either way — this PR touches docs plus one constant.

> [!NOTE]
> **Release notes are not in this PR.** The plan lists a CHANGELOG entry, but this repo has no CHANGELOG file — stable releases use a human-edited draft release. **Whoever publishes the release that carries #947 must call out the breaking change explicitly:** the `Serilog` configuration section is no longer read, and `ConfigureSerilog` / `ConfigureBootstrapLogger` are removed. It should not be left to the diff.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

Refs #476

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -428,9 +428,46 @@ Rules:
 
 ---
 
+## Observability
+
+Logs, metrics and traces come from **one mechanism**: the OpenTelemetry SDK, exporting OTLP. There is
+no second logging stack and no vendor SDK in the dependency graph.
+
+| Concern | Decision |
+|---|---|
+| Signals | All three from the OTel SDK, sharing one resource so a record and its span agree on `service.name` |
+| Activation | Off until an OTLP endpoint resolves. Nothing configured means no exporter allocated |
+| Precedence | `OTEL_*` environment variables are authoritative; `Endatix:Telemetry` is a fallback, never an override |
+| Registration | `EndatixTelemetryBuilder` owns all three exporters; `EndatixLoggingBuilder` owns the provider list |
+| Failure mode | A malformed endpoint or protocol throws at startup rather than exporting nothing quietly |
+| Vendors | None shipped. Consumers add their own through `Logging.Configure(...)` or `AddOpenTelemetry()` |
+
+**Why the log exporter lives in the telemetry builder, not the logging builder.** All three signals
+have to resolve the same endpoint, protocol and resource. Splitting that across two builders means
+duplicating the resolution and risking a log record whose `service.name` disagrees with its span —
+at which point nothing joins them.
+
+**Ordering between the two builders is load-bearing.** `EndatixLoggingBuilder` calls
+`ClearProviders()`, and `AddLogging` applies its callback immediately rather than at provider-build
+time. The OTel log provider therefore survives only because `Logging.UseDefaults()` runs before
+`Telemetry.Build()`, which `FinalizeConfiguration()` guarantees. A test pins this.
+
+**Serilog's remaining role.** It is not the logging pipeline. It is the rotation implementation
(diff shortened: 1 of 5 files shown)


PeterMarinov (approved):
Approved. 

Can you add notes to devs who are used to looking the Serilog output locally - it will now look in JSON ready for OTel Export (below):

Image

Can you suggest best doc experience e.g. start aspire or just looking for the dev persona, still using `dotnet run` locally

PR #963, opened 2026-08-19
feat(e960): Migrate smtp to the mailkit library

### LINKED ISSUE #960: Migrate SmtpEmailSender to MailKit/MimeKit
## Problem

`SmtpEmailSender` is built on `System.Net.Mail.SmtpClient`. Microsoft's API reference states that type
is not recommended for new development because it doesn't support many modern protocols, and points at
[MailKit](https://github.com/jstedfast/MailKit).

The concrete gap: `System.Net.Mail` **cannot do implicit TLS (SMTPS on port 465)** at all. Customers
whose relay requires it can't use our SMTP provider today. A customer has already sent us a working
MailKit-based replacement.

## Proposal

Swap the transport inside `SmtpEmailSender` from `System.Net.Mail` to `MailKit.Net.Smtp` + `MimeKit`.

**Not** a new email provider — `ProviderName` stays `"SMTP"`, the constructor, `IsConfigured`, DI
registration and the `Endatix:Integrations:Email:SmtpSettings` section are all unchanged. No existing
`appsettings.json` needs editing.

## Scope

- [ ] Add `MailKit` / `MimeKit` to `Directory.Packages.props` + `Endatix.Infrastructure.csproj`
- [ ] Rewrite `SmtpEmailSender` internals; delete the `System.Net.Mail` path
- [ ] Add `SmtpSettings.SecurityMode` (new `SmtpSecurityMode` enum, defaults to deriving from
      `EnableSsl`/`Port`) and `SmtpSettings.CheckCertificateRevocation` (defaults `false`)
- [ ] Add a `protected virtual CreateTransport()` seam — the send path currently has zero test coverage
- [ ] Tests: TLS mapping theory, connect/authenticate/send, blank username, multipart ordering, disconnect-on-throw
- [ ] Docs: settings reference + upgrade note

## Compatibility

Behaviour-preserving by default. Three deliberate deviations from a naive port:

1. **TLS mapping.** `EnableSsl → SslOnConnect` would break every existing deployment (defaults are
   port 587 + `EnableSsl: true`, which is STARTTLS today). Map to `StartTls`, using `SslOnConnect` only
   for port 465. `StartTls` rather than `Auto` — `Auto` degrades to plaintext, which would be a silent
   security downgrade.
2. **Certificate revocation.** MailKit checks by default, `System.Net.Mail` doesn't. Default to `false`
   to match current behaviour; firewalls blocking OCSP/CRL would otherwise fail the handshake closed.
3. **Blank username.** Currently means "use Windows process identity" (`UseDefaultCredentials`); MailKit
   can't replicate that. Becomes anonymous, with a one-time warning logged. No-op on Linux containers;
   a real break for Windows installs relaying through Exchange — call out in release notes.

No rollback switch — a real-relay smoke test (587+STARTTLS, 465+implicit TLS, anonymous relay) is
required before release.

## References

- [SmtpClient — Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient)

orilux:
## Description

Migrates `SmtpEmailSender` from `System.Net.Mail` to MailKit/MimeKit. Microsoft's own docs recommend against `SmtpClient` for new development, and the old transport couldn't do implicit TLS (port 465) at all. Behavior-preserving by default — same `SmtpSettings` section, same constructor and provider name, no config changes needed for existing STARTTLS deployments.

Two new opt-in settings (`SecurityMode`, `CheckCertificateRevocation`), one flagged behavior change (blank `Username` no longer uses Windows integrated auth — logs a warning instead, since MailKit has no equivalent).

Also fixes an unrelated bug found while testing: `TypedResultsBuilder`/`ResultExtensions` never handled `ResultStatus.Unavailable`, so "Send Test Email" crashed with an `InvalidCastException` instead of returning a clean 503 whenever the provider wasn't configured.

## Changes

- `SmtpEmailSender.cs` — MailKit transport, TLS mapping (`StartTls` default / `SslOnConnect` @465), `CheckCertificateRevocation` wiring, `CreateTransport()` testability seam
- `SmtpSettings.cs` / `SmtpSecurityMode.cs` — new settings, additive only
- `TypedResultsBuilder.cs` / `ResultExtensions.cs` — added missing `ResultStatus.Unavailable` → 503 handling
- `email-settings.mdx` — documents new settings + upgrade note
- Tests: TLS-mapping theory (11 cases), transport/behavior coverage, regression tests for the `Unavailable` fix

## Test plan

- [x] Full unit suite green (pre-existing, unrelated `EndatixFileLoggingTests` file-handle flake aside)
- [x] Verified against local Mailpit — emails delivered
- [x] "Send Test Email" returns a clean 503 when the provider is unconfigured
- [ ] Real-relay smoke test before release: 587+STARTTLS with auth (must not regress), 465+implicit TLS (new), anonymous relay

## Related Issues
closes #960 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Summary by CodeRabbit

- **New Features**
  - SMTP email delivery now supports MailKit with STARTTLS and implicit TLS/SMTPS.
  - Added configurable security modes and optional TLS certificate revocation checks.
  - HTML and plain-text email alternatives are preserved.

- **Bug Fixes**
  - Service-unavailable results now return HTTP 503 responses with appropriate problem details.

- **Documentation**
  - Updated SMTP configuration guidance, security options, authentication behavior, and upgrade notes.


src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -59,13 +65,28 @@
...
+        {
+            await ConnectAndAuthenticateAsync(smtpClient, cancellationToken);
+            await smtpClient.SendAsync(mimeMessage, cancellationToken);
 
-        _logger.LogInformation("SMTP email sent successfully to {To} with subject {Subject}",
-            SensitiveValue.Email(email.To), email.Subject);
+            // Logged here, not after the finally: a DisconnectAsync failure during cleanup must not
+            // suppress this record of a send that already succeeded.
+            _logger.LogInformation("SMTP email sent successfully to {To} with subject {Subject}",
+                SensitiveValue.Email(email.To), email.Subject);
```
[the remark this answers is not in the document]
orilux: False positive. `SensitiveValue.Email(email.To)` doesn't expose the address — `PiiRedactor.RedactEmail()` returns the hardcoded constant `"[redacted-email]"` regardless of input, so the raw value never reaches the log sink. CodeQL's taint tracker can't see through this bespoke sanitizer, hence the flag. This line is also unchanged from the pre-migration code and matches the same redaction pattern used elsewhere in the codebase (`SendVerificationEmailHandler`, `AppUserRegistrationService`). Dismissing as false positive.

tests/Endatix.Infrastructure.Tests/Email/SmtpEmailSenderTests.cs
```diff
@@ -55,4 +66,288 @@ await templateRepository.Received(1).FirstOrDefaultAsync(
...
+            HtmlBody = "<html>Hello World</html>"
+        };
+
+        // Act
+        await sut.SendEmailAsync(email, CancellationToken.None);
+
+        // Assert
+        logger.ReceivedCalls().Should().ContainSingle(call =>
+            call.GetMethodInfo().Name == nameof(ILogger.Log) &&
+            (LogLevel)call.GetArguments()[0]! == LogLevel.Warning);
```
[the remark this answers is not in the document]
orilux: Addressed

src/Endatix.Infrastructure/Email/SmtpSecurityMode.cs
```diff
@@ -0,0 +1,26 @@
...
+
+/// <summary>
+/// Controls how <see cref="SmtpEmailSender"/> secures its connection to the SMTP server.
+/// </summary>
+public enum SmtpSecurityMode
+{
+    /// <summary>
+    /// Derive the mode from <see cref="SmtpSettings.EnableSsl"/> and <see cref="SmtpSettings.Port"/>:
+    /// disabled when EnableSsl is false, implicit TLS on port 465, STARTTLS otherwise.
+    /// Reproduces the behavior of the pre-MailKit SmtpClient-based sender.
```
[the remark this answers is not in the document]
orilux: Addressed

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -59,13 +65,28 @@ public async Task SendEmailAsync(EmailWithBody email, CancellationToken cancella
...
+        }
+        finally
+        {
+            // CancellationToken.None: an already-cancelled token here would make the cleanup itself
+            // throw, masking whatever exception (if any) is already propagating from the try block.
+            if (smtpClient.IsConnected)
+            {
+                await smtpClient.DisconnectAsync(true, CancellationToken.None);
+            }
```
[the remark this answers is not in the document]
orilux: Addressed

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -59,13 +65,43 @@ public async Task SendEmailAsync(EmailWithBody email, CancellationToken cancella
...
+            try
+            {
+                // CancellationToken.None: an already-cancelled token here would make the cleanup
+                // itself throw, masking whatever exception is already propagating from the try block.
+                if (smtpClient.IsConnected)
+                {
+                    await smtpClient.DisconnectAsync(true, CancellationToken.None);
+                }
+            }
+            catch (Exception cleanupException) when (primaryException is not null)
```
igdragoev: **A cleanup failure on the success path escapes to the caller.**

The filter `when (primaryException is not null)` only swallows a `DisconnectAsync` failure when the send has *already* failed. On the success path the filter doesn't match, so the cleanup exception propagates out of `SendEmailAsync` — after the message was accepted with a 250 OK.

Scenario: the relay accepts the mail, then drops the connection or the TLS shutdown faults during QUIT. `SendVerificationEmailHandler` / `ResetPasswordHandler` / `AppUserRegistrationService` all see an exception, report failure, and a retry sends the same email a second time.

The comment above shows this was anticipated — the `LogInformation` was deliberately moved before the `finally` for exactly this reason — but the log line is the only thing protected; the caller isn't. Dropping the filter (and logging the cleanup failure in both cases) closes it:

```csharp
catch (Exception cleanupException)
{
    _logger.LogWarning(cleanupException, "SMTP disconnect failed.");
}
```

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -78,47 +114,76 @@ public async Task SendEmailAsync(EmailWithTemplate email, CancellationToken canc
...
+    protected virtual IMailTransport CreateTransport() => new SmtpClient();
+
+    private async Task ConnectAndAuthenticateAsync(IMailTransport smtpClient, CancellationToken cancellationToken)
     {
-        var smtpClient = new SmtpClient(_settings.Host, _settings.Port)
-        {
-            EnableSsl = _settings.EnableSsl,
-            UseDefaultCredentials = string.IsNullOrEmpty(_settings.Username)
-        };
+        smtpClient.CheckCertificateRevocation = _settings.CheckCertificateRevocation;
```
igdragoev, line 124: **MailKit does not honour `ServicePointManager.ServerCertificateValidationCallback`.**

`System.Net.Mail.SmtpClient` did. Any deployment pointing at an internal relay with a self-signed or otherwise untrusted certificate — which worked because of a global `ServicePointManager` override somewhere in startup — will now fail the TLS handshake instead of connecting.

The sender exposes no `ServerCertificateValidationCallback` setting and no seam to relax it; `CheckCertificateRevocation` is the only TLS knob. That makes it a real upgrade break with no configuration-level workaround.

It also isn't covered by the "Two behaviors changed" warning in `email-settings.mdx`. Either expose a setting for it, or document it as a third behavior change so operators hitting the handshake failure know why.

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -78,47 +114,76 @@ public async Task SendEmailAsync(EmailWithTemplate email, CancellationToken canc
...
+
+    private MimeMessage CreateMimeMessage(EmailWithBody email)
     {
         var fromAddress = string.IsNullOrEmpty(email.From) ? _settings.DefaultFromAddress : email.From;
-        var fromDisplayName = _settings.DefaultFromName;
 
-        var mailMessage = new MailMessage
+        var mimeMessage = new MimeMessage();
+        mimeMessage.From.Add(new MailboxAddress(_settings.DefaultFromName, fromAddress));
+        mimeMessage.To.Add(MailboxAddress.Parse(email.To));
```
igdragoev: **`MailboxAddress.Parse` accepts exactly one mailbox; the old call accepted a list.**

`MailMessage.To.Add(string)` documented support for a comma-separated address list. `MailboxAddress.Parse` throws `ParseException` on anything with more than one address, so a deployment or template that puts multiple recipients in `To` now fails instead of sending.

All in-repo callers pass a single address, so this is latent — but `To` is a free-form `string` on `BaseEmailModel`, so it's reachable from outside the repo. Preserving the old contract is a one-liner:

```csharp
mimeMessage.To.AddRange(InternetAddressList.Parse(email.To));
```

src/Endatix.Api/Infrastructure/TypedResultsBuilder.cs
```diff
@@ -182,6 +182,7 @@ public static implicit operator Results<Ok<TData>, ProblemHttpResult>(TypedResul
             ResultStatus.Unauthorized => resultMapper.SourceResult.ToProblem(resultMapper.ErrorMessage),
             ResultStatus.Forbidden => resultMapper.SourceResult.ToProblem(resultMapper.ErrorMessage),
             ResultStatus.Error => resultMapper.SourceResult.ToProblem(resultMapper.ErrorMessage),
+            ResultStatus.Unavailable => resultMapper.SourceResult.ToProblem(resultMapper.ErrorMessage),
             _ => throw new InvalidCastException($"Cannot cast the {resultMapper.SourceResult.ValueType} with status {resultMapper.SourceResult.Status}")
```
igdragoev, line 187: **`ResultStatus.CriticalError` still falls into this `throw` — the same failure this PR is fixing.**

`Unavailable` was added to both `ProblemHttpResult` operators, but `CriticalError` isn't mapped, so any handler returning `Result.CriticalError` through an `Ok<T>`/`ProblemHttpResult` endpoint crashes with an `InvalidCastException` → 500 rather than a problem response.

`ResultExtensions.ToEndpointResponse` (line 167) already maps `CriticalError`, so the two conversion paths now disagree about which statuses are representable. No handler produces it today, but adding the line here — and in the `Created<TData>` operator below — prevents the next occurrence of exactly this bug.

src/Endatix.Api/Infrastructure/ResultExtensions.cs
```diff
@@ -92,6 +93,7 @@ public static ProblemHttpResult ToProblem(this AppDomain.IResult result, string?
             ResultStatus.Unauthorized => (StatusCodes.Status401Unauthorized, ResultTitles.UNAUTHORIZED),
             ResultStatus.Forbidden => (StatusCodes.Status403Forbidden, ResultTitles.FORBIDDEN),
             ResultStatus.Error => (StatusCodes.Status500InternalServerError, ResultTitles.INTERNAL_SERVER_ERROR),
+            ResultStatus.Unavailable => (StatusCodes.Status503ServiceUnavailable, ResultTitles.SERVICE_UNAVAILABLE),
```
igdragoev, line 96: **503 is now reachable but undeclared on the endpoint that returns it.**

Commenting here since `SendTestEmail.cs` isn't in this diff — the issue is at `src/Endatix.Api/Endpoints/Admin/Email/SendTestEmail.cs:38`:

```csharp
.Produces<string>(200, "application/json")
.ProducesProblem(400));
```

With this mapping in place, that endpoint can now return 503, but the OpenAPI document won't model it — and `api-contracts` generates the Kotlin / .NET / Orval clients from that document, so callers get no typed handling for the new status. Worth adding `.ProducesProblem(503)` plus an `s.Responses[503]` summary alongside the existing 200/400 entries.

src/Endatix.Api/Infrastructure/ResultExtensions.cs
```diff
@@ -18,6 +18,7 @@ public static class ResultTitles
         public const string CONFLICT = "There was a conflict";
         public const string BAD_REQUEST = "There was a problem with your request";
         public const string INTERNAL_SERVER_ERROR = "An unexpected error occurred";
+        public const string SERVICE_UNAVAILABLE = "Service unavailable.";
```
igdragoev: **Duplicates a literal that already exists, and breaks the title convention.**

`"Service unavailable."` is already hardcoded at `ResultExtensions.cs:271` in `UnavailableEntity`. With this constant added, the same title is emitted from two independent sources, so they can drift — `UnavailableEntity` should reference `ResultTitles.SERVICE_UNAVAILABLE`.

Separately, the trailing period is inconsistent with every other value in this class (`"Resource not found"`, `"Unauthorized access"`, `"There was a conflict"`, …). These titles are part of the response contract, so it's worth settling now rather than after clients start matching on it.

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
```diff
@@ -78,47 +114,76 @@ public async Task SendEmailAsync(EmailWithTemplate email, CancellationToken canc
...
+    /// unencrypted connection when the server doesn't advertise STARTTLS, whereas EnableSsl = true
+    /// previously guaranteed encryption or failure.
+    /// </summary>
+    internal SecureSocketOptions ResolveSecureSocketOptions() => _settings.SecurityMode switch
+    {
+        SmtpSecurityMode.None => SecureSocketOptions.None,
+        SmtpSecurityMode.StartTls => SecureSocketOptions.StartTls,
+        SmtpSecurityMode.StartTlsWhenAvailable => SecureSocketOptions.StartTlsWhenAvailable,
+        SmtpSecurityMode.SslOnConnect => SecureSocketOptions.SslOnConnect,
+        _ => !_settings.EnableSsl
```
igdragoev, line 181: **`Port: 465` + `EnableSsl: false` resolves to `None`, which hangs rather than errors.**

In the `Auto` branch, `!_settings.EnableSsl` short-circuits before the port check, so that combination yields `SecureSocketOptions.None`. A plaintext SMTP client against an implicit-TLS listener typically blocks waiting for a greeting that never arrives, so the symptom is a hang until socket timeout rather than a clean failure — considerably harder to diagnose than a connection error.

The behavior is deliberate and covered by `InlineData(465, false, Auto, None)`, so this is a design question rather than a defect: a user who reads `EnableSsl` as "STARTTLS on/off" lands here without realising it. Either let the 465 heuristic win over `EnableSsl`, or log a warning when the combination is seen.

orilux: Addressed - went with logging unconditionally rather than letting port 465 override `EnableSsl`. Two reasons, both tied to how the rest of this migration was designed: `EnableSsl: false` staying unconditional (even at 465) is what actually matches `System.Net.Mail`'s old behavior — it was always plaintext regardless of port, so nothing here changes that. And `SecurityMode: SslOnConnect` already exists as an explicit escape hatch for anyone who deliberately wants port 465 to force TLS, so the fix doesn't need to live in the heuristic itself. Silently letting one setting override another explicit one is exactly the kind of surprise the `Auto` → `StartTls` (not `MailKit.SecureSocketOptions.Auto`) decision elsewhere in this PR was built to avoid.

The warning is process-wide/log-once (`Interlocked`-gated, same pattern as the blank-username warning), fires only when `SecurityMode` is left at `Auto` — an explicit `SecurityMode` is a deliberate choice and isn't second-guessed — and names the fix directly: set `EnableSsl: true` or `SecurityMode: SslOnConnect`.

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
orilux: Addressed. The `try/finally` no longer masks a real success as a failure: removed the `when (primaryException is not null)` guard entirely and simplified away the now-unneeded `primaryException` tracking, since an unconditional catch-and-log in the cleanup handles all four scenarios correctly on its own.

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
orilux: Addressed.  `MailboxAddress.Parse` → `InternetAddressList.AddRange(...Parse(...))`, restoring the old multi-recipient contract.

src/Endatix.Api/Infrastructure/TypedResultsBuilder.cs
orilux: Addressed. `ResultStatus.CriticalError` now routes through `ToProblem()` in both `ProblemHttpResult` operators, same pattern as `Unavailable`.

src/Endatix.Api/Infrastructure/ResultExtensions.cs
orilux: Addressed.  `SendTestEmail.cs` now declares `.ProducesProblem(503)` + a `Responses[503]` summary.

src/Endatix.Api/Infrastructure/ResultExtensions.cs
orilux: Addressed.  `UnavailableEntity` now references `ResultTitles.SERVICE_UNAVAILABLE` instead of its own hardcoded literal; trailing period removed to match the class's convention.

src/Endatix.Infrastructure/Email/SmtpEmailSender.cs
orilux: Agree - verified `ServerCertificateValidationCallback` is never touched anywhere in the sender, and confirmed the upgrade note doesn't cover it.

Going with documenting this as a third behavior change rather than adding a bypass setting. The two aren't quite symmetric: `CheckCertificateRevocation: false` was safe to ship because it preserves old behavior (revocation was never checked before either), whereas a validation-bypass flag would be new capability, not parity — and a JSON boolean is arguably easier to misconfigure into than the old global `ServicePointManager` override ever was.

The actual fix doesn't need code from us: MailKit's default validation checks the *machine trust store*, so installing the internal/self-signed CA there resolves the handshake failure with no bypass required — and is more secure than the old pattern, which typically disabled validation process-wide rather than just for SMTP. I'll add that as the third bullet in the upgrade note. If a deployment genuinely can't add to the trust store, a proper override is worth designing as its own follow-up rather than a rider on this PR.

igdragoev (approved):

Issue #964, opened 2026-08-19
Add Data Lists management list and items search API


PeterMarinov:
### Goal
Add authenticated management endpoints so Hub can search and page Data Lists and their items without loading full catalogs into memory.

### Implementation Notes
- **API-1:** `GET /data-lists` optional `query` (name/description contains) alongside existing `hasLocale`, `page`, `pageSize`.
- **API-2:** `GET /data-lists/{id}/items` with `query`, `matchMode`, `locale`, `includeLocales`, paging; `GET /data-lists/{id}?includeItems=false` for metadata-only details.
- **Out of scope:** per-item POST/PATCH/DELETE (#809).
- Related epic: #937
- Stacked PRs: API-1 → API-2 (see `.cursor/plans/h704-data-lists-grids-translations-plan.md` in endatix-saas)

PR #965, opened 2026-08-19
feat(e964): add list query search for management grids

PeterMarinov:
## Summary
- Add optional `search` on `GET /data-lists` (case-insensitive name/description contains) using `ISearchablePagedRequest` + `SearchablePagedRequestValidator`, same contract as other management lists.
- Keep paging, `hasLocale` (valid culture when set), and `Forms.View`.
- Part of stacked delivery for endatix/endatix#964 (API-1).

## Test plan
- [x] `dotnet test` ListTests + ListDataListsHandlerTests
- [ ] OpenAPI: `search` (not `query`) on List
- [ ] Hub: pass `search` on the data-lists grid request

## Summary by CodeRabbit

* **New Features**
  * Added optional search support when listing data lists.
  * Search matches data-list names and descriptions, without regard to letter casing.
  * Added locale filtering validation and documented the new search option in the endpoint.
* **Bug Fixes**
  * Improved total-result counts so they reflect active search filters.
* **Tests**
  * Added coverage for search forwarding, matching, and non-matching results.

### DIFF
--- a/src/Endatix.Api/Endpoints/DataLists/List.cs
+++ b/src/Endatix.Api/Endpoints/DataLists/List.cs
@@ -1,9 +1,12 @@
 using Endatix.Api.Common;
+using Endatix.Api.Endpoints.Common;
 using Endatix.Api.Infrastructure;
 using Endatix.Core.Abstractions.Authorization;
+using Endatix.Core.Infrastructure.Paging;
 using Endatix.Core.Infrastructure.Result;
 using Endatix.Core.UseCases.DataLists.List;
 using FastEndpoints;
+using FluentValidation;
 using MediatR;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.HttpResults;
@@ -27,11 +30,12 @@ public sealed class List(
         Summary(s =>
         {
             s.Summary = "List data lists";
-            s.Description = "Lists data lists for the current tenant with paging and an optional locale filter.";
+            s.Description = "Lists data lists for the current tenant with paging, optional name/description search, and an optional locale filter.";
             s.ExampleRequest = new DataListsListRequest
             {
                 Page = 1,
                 PageSize = 20,
+                Query = "cities",
                 HasLocale = "es"
             };
             s.ResponseExamples[200] = new Paged<DataListModel>(
@@ -64,7 +68,7 @@ public sealed class List(
     /// <inheritdoc />
     public override async Task<Results<Ok<Paged<DataListModel>>, ProblemHttpResult>> ExecuteAsync(DataListsListRequest request, CancellationToken ct)
     {
-        var result = await mediator.Send(new ListDataListsQuery(request.Page, request.PageSize, request.HasLocale), ct);
+        var result = await mediator.Send(new ListDataListsQuery(request.Page, request.PageSize, request.HasLocale, request.Query), ct);
 
         if (!result.IsSuccess)
         {
@@ -88,6 +92,12 @@ public sealed class DataListsListValidator : Validator<DataListsListRequest>
     public DataListsListValidator()
     {
         Include(new PageableRequestValidator());
+        RuleFor(x => x.Query)
+            .MaximumLength(PagedRequestLimits.MAX_SEARCH_LENGTH)
+            .When(x => !string.IsNullOrWhiteSpace(x.Query));
+        RuleFor(x => x.HasLocale)
+            .IsCultureCode()
+            .When(x => !string.IsNullOrWhiteSpace(x.HasLocale));
     }
 }
 
@@ -106,4 +116,9 @@ public sealed class DataListsListRequest : IPagedRequest
     /// Optional locale code; returns only lists whose AvailableLocales contain this code.
     /// </summary>
     public string? HasLocale { get; set; }
+
+    /// <summary>
+    /// Optional name/description search (case-insensitive contains).
+    /// </summary>
+    public string? Query { get; set; }
 }
--- a/src/Endatix.Api/Endpoints/DataLists/List.cs
+++ b/src/Endatix.Api/Endpoints/DataLists/List.cs
@@ -2,7 +2,6 @@ using Endatix.Api.Common;
 using Endatix.Api.Endpoints.Common;
 using Endatix.Api.Infrastructure;
 using Endatix.Core.Abstractions.Authorization;
-using Endatix.Core.Infrastructure.Paging;
 using Endatix.Core.Infrastructure.Result;
 using Endatix.Core.UseCases.DataLists.List;
 using FastEndpoints;
@@ -35,7 +34,7 @@ public sealed class List(
             {
                 Page = 1,
                 PageSize = 20,
-                Query = "cities",
+                Search = "cities",
                 HasLocale = "es"
             };
             s.ResponseExamples[200] = new Paged<DataListModel>(
@@ -68,7 +67,7 @@ public sealed class List(
     /// <inheritdoc />
     public override async Task<Results<Ok<Paged<DataListModel>>, ProblemHttpResult>> ExecuteAsync(DataListsListRequest request, CancellationToken ct)
     {
-        var result = await mediator.Send(new ListDataListsQuery(request.Page, request.PageSize, request.HasLocale, request.Query), ct);
+        var result = await mediator.Send(new ListDataListsQuery(request.Page, request.PageSize, request.HasLocale, request.Search), ct);
 
         if (!result.IsSuccess)
         {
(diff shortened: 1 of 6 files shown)


tests/Endatix.Core.Tests/UseCases/DataLists/List/ListDataListsHandlerTests.cs
```diff
@@ -134,4 +158,16 @@ bool Matches(DataList dataList) =>
...
+    private static bool SpecFiltersByQuery(ISpecification<DataList> spec, string query)
+    {
+        DataList matching = new(SampleData.TENANT_ID, "Cities", "Major cities");
+        DataList other = new(SampleData.TENANT_ID, "Countries", "ISO codes");
+
+        bool Matches(DataList dataList) =>
+            spec.WhereExpressions.Any()
+            && spec.WhereExpressions.All(expression => expression.FilterFunc(dataList));
+
+        return Matches(matching) && !Matches(other);
```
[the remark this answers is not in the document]
PeterMarinov: Fix added - check and valdiate

orilux (approved):
Approved

src/Endatix.Api/Endpoints/DataLists/List.cs
```diff
@@ -87,21 +90,27 @@ public sealed class DataListsListValidator : Validator<DataListsListRequest>
     /// </summary>
     public DataListsListValidator()
     {
-        Include(new PageableRequestValidator());
+        Include(new SearchablePagedRequestValidator());
```
orilux: Swapping `PageableRequestValidator` → `SearchablePagedRequestValidator` doesn't just add the search-length rule — it also brings `PageSize <= PagedRequestLimits.MAX_PAGE_SIZE` (100). Combined with the new `IsCultureCode()` rule below, two request shapes that returned `200` before this PR now return `400`:

 - `GET /data-lists?pageSize=500` → was served (`PagingParameters` doesn't cap page size, so this really did materialize 500 rows); now rejected.
 - `GET /data-lists?hasLocale=default` (or any non-culture string) → was `200` with an empty page; now rejected.

 Both are improvements — the page-size cap in particular closes a real unbounded-fetch hole. But this is a contract change on a shipped endpoint and the PR body reads as additive ("Keep paging, hasLocale"). Please call it out in the release notes.

 Hub check: `get-data-lists.action.ts:19` sends `pageSize = 100` in a fetch-all loop — passes, but sits exactly on the new cap, so any future bump breaks. It never sends `hasLocale`, so that rule is safe today.

src/Endatix.Api/Endpoints/DataLists/List.cs
```diff
@@ -87,21 +90,27 @@ public sealed class DataListsListValidator : Validator<DataListsListRequest>
...
+        RuleFor(x => x.HasLocale)
```
orilux: This PR adds two new rejection paths (`HasLocale` must be a real culture; `PageSize <= 100`) with no coverage. Every other DataLists endpoint that has a validator has a matching test file - `AddDataListLocaleValidatorTests`, `RemoveDataListLocaleValidatorTests`, `SetDataListDefaultLocaleValidatorTests`, `ListDataListItemsValidatorTests`, `ImportDataListValidatorTests`, `ExportDataListValidatorTests`.

 A `ListDataListsValidatorTests.cs` following `ListDataListItemsValidatorTests` would cover it in ~20 lines: valid request passes; `HasLocale = "default"` / `"not-a-culture"` errors; `PageSize = 101` errors; `HasLocale = null` and `HasLocale = ""` both pass (the optional-filter contract).

tests/Endatix.Core.Tests/UseCases/DataLists/List/ListDataListsHandlerTests.cs
```diff
@@ -134,4 +158,22 @@ bool Matches(DataList dataList) =>
 
         return Matches(withLocale) && !Matches(withoutLocale);
     }
+
+    private static bool SpecFiltersBySearch(ISpecification<DataList> spec, string search)
+    {
+        var normalizedTerm = search.Trim().ToLowerInvariant();
```
orilux: `SpecFiltersBySearch` re-derives the production normalization (`search.Trim().ToLowerInvariant()`) to build its fixtures, so any normalization bug is applied to both sides and cancels out. The assertion genuinely does cover case-insensitivity and the name-OR-description branch — I checked each mutation — but drop `.Trim()` from `ApplyListFilters` and this test still passes, because the input has no surrounding whitespace.

Cheapest fix: change the input at line 113 to `Search: " MaJoR "`. The helper already normalizes to `"major"`, so the fixtures are unchanged and the test starts failing if `Trim()` disappears.

tests/Endatix.Core.Tests/UseCases/DataLists/List/ListDataListsHandlerTests.cs
```diff
@@ -97,6 +97,30 @@ await _repository.Received(1).ListAsync(
...
+        _repository.CountAsync(Arg.Any<DataListsSpecifications.ListSpec>(), Arg.Any<CancellationToken>())
+            .Returns(1);
+        _repository.ListAsync(Arg.Any<DataListsSpecifications.ListWithPagingToDtoSpec>(), Arg.Any<CancellationToken>())
+            .Returns(new List<DataListDto>
+            {
+                new(5, "Cities", "Major cities", DateTime.UtcNow, null, true, 1, "en", [], Array.Empty<DataListItemDto>())
+            });
+
+        // Act
+        await _sut.Handle(new ListDataListsQuery(1, 10, Search: "MaJoR"), TestContext.Current.CancellationToken);
```
orilux: `SpecFiltersByLocale` and `SpecFiltersBySearch` both assert over `WhereExpressions.All(...)`, so neither can tell which expression is which — they only work because exactly one filter is set per test. Now that `ApplyListFilters` builds both, a case with `hasLocale` *and* `search` set would be worth adding, asserting a list that matches the locale but not the term is excluded. That's the combination the Hub grid will actually send.

src/Endatix.Core/Specifications/DataListsSpecifications.cs
```diff
@@ -63,6 +55,26 @@ public ListWithPagingToDtoSpec(PagingParameters pagingParams, string? hasLocale
...
+        if (!string.IsNullOrWhiteSpace(hasLocale))
+        {
+            var locale = hasLocale.Trim().ToLowerInvariant();
+            query.Where(x => x.AvailableLocales.Contains(locale));
+        }
+
+        if (!string.IsNullOrWhiteSpace(search))
+        {
+            var term = search.Trim().ToLowerInvariant();
+            query.Where(x =>
```
orilux: Worth noting explicitly since the PR body says "same contract as other management lists": `FormsListFilterSpec.ApplyNameSearch` matches **name only**, while this matches **name or description**. Same query param, different semantics. Not wrong — description search is more useful here — but the Hub's search placeholder and the API docs should say which fields are covered, otherwise users will assume the Forms behavior.

 (The `x.Description != null` guard is redundant in SQL — `LOWER(NULL) LIKE …` is already `NULL`/false — but it's needed for the in-memory `FilterFunc` evaluation in the tests, so leave it.)

src/Endatix.Core/Specifications/DataListsSpecifications.cs
```diff
@@ -63,6 +55,26 @@ public ListWithPagingToDtoSpec(PagingParameters pagingParams, string? hasLocale
...
+        string? hasLocale,
+        string? search)
+    {
+        if (!string.IsNullOrWhiteSpace(hasLocale))
+        {
+            var locale = hasLocale.Trim().ToLowerInvariant();
+            query.Where(x => x.AvailableLocales.Contains(locale));
+        }
+
+        if (!string.IsNullOrWhiteSpace(search))
```
orilux: `LOWER(col) LIKE '%term%'` is non-sargable, and the handler runs the filter twice per request - once for `CountAsync`, once for the page - so that's two full scans over two columns. Almost certainly fine at data-list volumes, and it matches `FormsListFilterSpec`, so no change requested here.

Flagging only because the repo already has a provider-aware path for this (`IRelationalSubstringLikeFilter` → Npgsql `ILIKE` / SQL Server `LIKE`, used by the data-list *items* search). It lives in `Endatix.Infrastructure`, so a Core spec can't reach it — if list search ever grows past small tenants, that's the seam to reuse rather than a third hand-rolled `ToLower().Contains()`.

src/Endatix.Api/Endpoints/DataLists/List.cs
```diff
@@ -87,21 +90,27 @@ public sealed class DataListsListValidator : Validator<DataListsListRequest>
...
+public sealed class DataListsListRequest : ISearchablePagedRequest
 {
     /// <inheritdoc />
     public int? Page { get; set; }
 
     /// <inheritdoc />
     public int? PageSize { get; set; }
 
+    /// <inheritdoc />
+    public string? Search { get; set; }
```
orilux: `GET /data-lists` now takes `search`; `GET /data-lists/{dataListId}/items` takes `query` (`ListDataListItemsRequest.Query`). Same feature area, two names for the same idea — the Hub will be passing both from adjacent screens. The test plan already flags this ("`search` (not `query`)"), so just confirming it's a deliberate split and worth a follow-up issue to converge, rather than something to fix in this PR.

Also, `Search` inherits its OpenAPI description from `ISearchableRequest` ("Optional free-text search term"), which doesn't mention name/description. The endpoint `s.Description` covers it, so this is fine as-is.

PR #966, opened 2026-08-19
feat(e964): Add management GET items endpoint and includeItems flag

PeterMarinov:
## Summary
- Add authenticated `GET /data-lists/{id}/items` for paged management item search (`query`/`matchMode`/`locale`/`includeLocales`, default `pageSize` 25); inactive lists included via `RequireActive=false`
- Add `includeItems` on `GET /data-lists/{id}` (default `true`) so Hub can load catalog metadata without item rows
- Keep items `query` (not `ISearchablePagedRequest.Search`) to stay aligned with public runtime search
- Refactor: reduce number of contructor params in `SearchDataListItemsQuery` and reuse that for related Commands
- Docs: clarify parent-list vs item-search params; OpenAPI example for metadata-only get-by-id

## Test plan
- [x] ListItemsTests (paging → skip/take, `RequireActive=false`, defaults)
- [x] ListDataListItemsValidatorTests
- [x] GetByIdTests (`includeItems` false + default true)
- [x] GetDataListByIdHandlerTests (`ByIdWithoutItemsToDtoSpec`)
- [ ] Hub stack (endatix-hub#865) after this lands

## Test plan
- [x] ListItemsTests + GetDataListByIdHandlerTests
- [ ] Hub stack (endatix-hub#865) should merge after this lands

### DIFF
--- a/docs/endatix-docs/docs/developers/api/data-lists-translations.mdx
+++ b/docs/endatix-docs/docs/developers/api/data-lists-translations.mdx
@@ -11,7 +11,7 @@ Data list choice labels are stored as a JSON map (`Labels`) with a synthetic `de
 
 ## Management list and items grids
 
-Authenticated Hub grids use the same search/paging contract as other list endpoints (`query` + `page` / `pageSize`). Hub maps its URL `search` param to `query`.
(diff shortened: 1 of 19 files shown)


orilux (approved):
Approved

PR #1010, opened 2026-09-03
fix(e1009): return 409 response when  assuming current tenant

### LINKED ISSUE #1009: Reject assume-tenant when the session is already in that tenant
### Goal
Reject `POST /auth/assume-tenant` when the session JWT `tid` already matches the target tenant — do not mint a new assumed session for a tenant the caller is already in.

### Implementation Notes
- Compare `request.TenantId` to session `tid` (`IUserContext.GetCurrentUser().TenantId`), not `AppUser.TenantId` from the database.
- **409 Conflict** — current session state already has that tenant. Not 400 (400 stays for nested assume / invalid id).
- Nesting (`act` set, different `tid`) stays **400** “Exit the current assumed tenant session first.” If `act` is set and `tid` equals the target, prefer **409** (already in that tenant).
- No tokens, no `tenant.context.changed`.

PeterMarinov:
# fix(e1009): return 409 response when  assuming current tenant

## Related Issues
- fixes #1009 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
image


src/Endatix.Core/UseCases/Identity/AssumeTenant/AssumeTenantHandler.cs
```diff
@@ -31,6 +31,11 @@ public async Task<Result<AuthTokensDto>> Handle(AssumeTenantCommand request, Can
             return Result.Unauthorized();
         }
 
+        if (actor.TenantId == request.TenantId)
```
igdragoev, line 36: If we have used this API in the HUB we have introduced a breaking change. You can ignore this comment if we are not using this endpoint yet.

**Client break isn't paired with a Hub change.** This flips a previously-successful call: `POST /auth/assume-tenant` with the caller's own tenant used to return 200 and a fresh assumed session, and now returns 409.

The Hub's tenant-switcher path handles 200/400/404 for this call, so a platform admin clicking their own tenant — or a Hub that uniformly calls assume-tenant on entering tenant view — now gets an unhandled error instead of navigating. Nothing in `endatix-hub` is referenced by this PR and no follow-up issue is linked, so the break would ship unannounced.

Worth either filing the Hub issue before merge, or confirming in the PR description that the Hub never calls assume-tenant with the current `tid`.

igdragoev (commented):
The fix itself matches #1009 exactly: it compares `request.TenantId` to the JWT-derived `IUserContext.GetCurrentUser().TenantId` (not `AppUser.TenantId`), orders the 409 ahead of the nesting 400, and returns before any token/event/cache work. `Conflict → 409` is verified through `ResultExtensions.ToProblem`, and `docs/endatix-docs/swagger.json` doesn't carry this endpoint, so no spec regeneration is needed.

Seven comments below — the first four are contract/correctness, the last three are test-strength and hygiene. (Note the SonarQube 0.0% new-code coverage is a non-signal here; coverage isn't uploaded for this project.)

tests/Endatix.Core.Tests/UseCases/Identity/AssumeTenant/AssumeTenantHandlerTests.cs
```diff
@@ -142,9 +175,9 @@ await _authorizationService.Received(1).InvalidateAuthorizationDataCacheAsync(
             ActorId.ToString(), TargetTenantId, Arg.Any<CancellationToken>());
     }
 
-    private User ArrangeActor()
+    private User ArrangeActor(long sessionTenantId = HomeTenantId)
```
igdragoev: **Neither new test can actually distinguish session `tid` from `AppUser.TenantId`.** `ArrangeActor` builds one `User` with `sessionTenantId` and hands the same instance to both `_userContext.GetCurrentUser()` and `_userService.GetUserAsync()` (line 180).

That is the exact distinction #1009 mandates ("compare to session `tid`, not `AppUser.TenantId` from the database"). A refactor to `userResult.Value.TenantId == request.TenantId` would keep both new tests green, while in production it would 409 an admin whose home tenant is the target but whose session is assumed elsewhere, and miss the real already-in-tenant case.

Suggest `ArrangeActor(long sessionTenantId, long homeTenantId)` with two distinct `User` instances, and a test where the two differ.

ARCHITECTURE.md
```diff
@@ -491,15 +490,15 @@ Gated by the deployment flag `multi-tenancy` (`FeatureFlags.MultiTenancy`). Off
...
+`POST /auth/assume-tenant` mints the assumed session, `POST /auth/exit-assume` returns to the home tenant. Both require `AuthorizationPolicies.PlatformAdminAccess`, sit behind `MultiTenancyGate` (off → 404), and answer with `TenantSessionResponse`. Assumed access tokens expire after `AssumeTenantSession.AccessExpiryMinutes`; refresh keeps that shorter lifetime while `act` is present. Sessions do **not** nest — assuming from an assumed session into a **different** tenant is a **400**; exit first. Assuming the tenant the session `tid` already has is a **409** (no new tokens). Neither handler writes membership; the audit trail is `tenant.context.changed` raised on the assumed tenant. Refresh persistence failure is returned as an infrastructure error (same as login), not a silent success.
```
igdragoev: **The mirror operation is left inconsistent with the rule being documented here.** `ExitAssumeHandler.cs:36` still returns `Result.Invalid(new ValidationError("Not in an assumed tenant session."))` → 400 for the same class of condition this PR reclassifies as 409: the session state already matches what the request asks for.

A client reading this sentence will reasonably branch on 409 for both no-op cases and mis-handle exit-assume. Either move exit-assume to 409 in this PR, or add a clause here saying why the two differ.

src/Endatix.Api/Endpoints/Auth/AssumeTenant.cs
```diff
@@ -29,16 +29,18 @@ public override void Configure()
             s.Description = "Switches the current PlatformAdmin session into a target tenant without impersonating a user or creating membership.";
             s.ExampleRequest = new AssumeTenantRequest { TenantId = 42 };
             s.Responses[200] = "Assumed session issued.";
-            s.Responses[400] = "Invalid tenant id, or the session is already assumed.";
+            s.Responses[400] = "Invalid tenant id, or the session is already assumed into a different tenant.";
             s.Responses[403] = "The current user is not a platform administrator.";
             s.Responses[404] = "Multi-tenancy is disabled, or the tenant was not found.";
+            s.Responses[409] = "The session is already in the target tenant.";
```
igdragoev, line 36: **While this list is being edited — it still omits 401.** `AssumeTenantHandler` returns `Result.Unauthorized()` when there is no current user, and `PlatformAdminAccess` rejects an unauthenticated caller with 401, but neither `s.Responses[401]` nor `.ProducesProblem(...Status401Unauthorized)` exists.

AGENTS.md requires the `Description(...)` list to match exactly the statuses the `Summary` declares, and the generated clients in `api-contracts` (Orval / Kotlin / .NET) get no 401 branch as a result. This block is the place to fix it.

tests/Endatix.Core.Tests/UseCases/Identity/AssumeTenant/AssumeTenantHandlerTests.cs
```diff
@@ -53,6 +53,39 @@ public async Task Handle_Anonymous_ReturnsUnauthorized()
...
+        // Arrange
+        ArrangeActor(sessionTenantId: TargetTenantId);
+
+        // Act
+        var result = await _sut.Handle(new AssumeTenantCommand(TargetTenantId), CancellationToken.None);
+
+        // Assert
+        result.Status.Should().Be(ResultStatus.Conflict);
+        result.Errors.Should().Contain("You cannot assume a tenant you are already in.");
+        _tokenService.DidNotReceive().IssueRefreshToken();
```
igdragoev: **`WithoutIssuingTokens` doesn't assert the whole claim.** #1009 says "no tokens, no `tenant.context.changed`", but only `IssueRefreshToken` and the tenant read are checked. Missing:

- `_tokenService.DidNotReceive().IssueAccessToken(...)`
- `_tenantRepository.DidNotReceive().UpdateAsync(...)` — the path that raises `tenant.context.changed`
- `_authorizationService.DidNotReceive().InvalidateAuthorizationDataCacheAsync(...)`

If the guard were later moved below the token-minting block (say, to reuse the loaded tenant), an access token would be issued and the authorization cache invalidated on a 409 path, and this test would still pass — `IssueRefreshToken` happens to run after `IssueAccessToken`.

src/Endatix.Core/UseCases/Identity/AssumeTenant/AssumeTenantHandler.cs
```diff
@@ -31,6 +31,11 @@ public async Task<Result<AuthTokensDto>> Handle(AssumeTenantCommand request, Can
...
+        {
+            return Result<AuthTokensDto>.Conflict("You cannot assume a tenant you are already in.");
```
igdragoev: **Inline conflict message, against the house convention.** Every other Conflict site names its message in a `private const`: `PartialUpdateFormHandler.DISABLE_CONFLICT_MESSAGE`, `UpdateFormHandler.DISABLE_CONFLICT_MESSAGE`, `CreateSubmissionHandler.DUPLICATE_CONFLICT_MESSAGE`, `ExternalAppUserProvisioner.DUPLICATE_EMAIL_MESSAGE`, `FolderAssignmentPolicy.IMMUTABLE_FOLDER_MOVE_ERROR`.

Here the literal is also repeated verbatim in `AssumeTenantHandlerTests.cs:67` and `AssumeTenantTests.cs:69`, so a reword touches three files.

ARCHITECTURE.md
```diff
@@ -491,15 +490,15 @@ Gated by the deployment flag `multi-tenancy` (`FeatureFlags.MultiTenancy`). Off
 
 **JWT session** (`AccessTokenIssueOptions` / `AccessTokenSession`). Refresh remints from the **token** (`tid` + optional `act`), not from `AppUser.TenantId`. One mint path; `act` is not impersonation (`sub` stays the admin). Outbox `tenant.context.changed` with kinds `assumed` / `exited`. Kind `switched` is reserved for a future membership product.
 
-| Session | `sub` | `tid` | `act` | Status |
-|---------|-------|-------|-------|--------|
-| Login | member | home tenant | absent | refresh keeps session `tid` |
-| Assume tenant | **admin** | target | admin | not impersonation |
-| On-behalf / login-as-customer | **customer** | customer tenant | support plus a **new claim** (do not reuse `act` alone) | **not implemented** |
+| Session                       | `sub`        | `tid`           | `act`                                                   | Status                      |
```
igdragoev: **Prettier re-alignment churn mixed into a one-line fix.** Of the ARCHITECTURE.md hunks, only line 501 is substantive — this table, the decision-log rows at 527-529, the trailing-space change at 332 and the blank-line removal at 340 are whitespace-only.

It makes the one meaningful line hard to find, and re-padded rows nobody edited create conflicts for the other long-lived architecture branches touching this file. Suggest reverting the whitespace-only hunks.

src/Endatix.Core/UseCases/Identity/AssumeTenant/AssumeTenantHandler.cs
PeterMarinov: Not a Hub pairing gap. Assume-tenant isn’t on Hub main yet (stacked https://github.com/endatix/endatix-hub/pull/903). That path maps API failures through toResult - 409 is a toast like 400/404, not an unhandled status (check PR screenshot showing that error).

Calling assume with the current session tid is exactly the case this PR forbids; we don’t need a Hub change before merge.

tests/Endatix.Core.Tests/UseCases/Identity/AssumeTenant/AssumeTenantHandlerTests.cs
PeterMarinov: Agreed - tests weren’t proving the session-vs-home distinction. ArrangeActor now stubs JWT tid and GetUserAsync separately, plus a case where home tenant == target but session tid does not (must not 409). Production still compares IUserContext.GetCurrentUser().TenantId.

ARCHITECTURE.md
PeterMarinov: Won’t map exit-assume to 409. That’s “not in an assumed session” (act missing), not “already in the tenant you asked for.” Documented the split in ARCHITECTURE.md so clients don’t treat both no-ops as 409

src/Endatix.Api/Endpoints/Auth/AssumeTenant.cs
PeterMarinov: Done: 401 on assume Summary + ProducesProblem. My main concern is that this applies to all endpoints that are auth protected and this is the only one that has the 401 officially documented. Maybe we should add shared docs for all endpoints so protected endpoints are less noisy

tests/Endatix.Core.Tests/UseCases/Identity/AssumeTenant/AssumeTenantHandlerTests.cs
PeterMarinov: Vald one. Conflict tests now assert no access-token mint, no tenant load/update (tenant.context.changed), and no authz cache invalidation — not only IssueRefreshToken. Fixed

src/Endatix.Core/UseCases/Identity/AssumeTenant/AssumeTenantHandler.cs
PeterMarinov: Pulled the 409 copy into AssumeTenantHandler.AlreadyInTenantMessage and pointed Core + Api tests at it.

ARCHITECTURE.md
PeterMarinov: Reverted the whitespace-only ARCHITECTURE.md hunks. Diff vs main is that one JWT-session paragraph (409 + why exit stays 400).

igdragoev (approved):

PR #1013, opened 2026-09-04
feat(helm): config passthrough — extraEnv, extraEnvFrom, volumes, podAnnotations, optional initialUser

igdragoev:
## Description

The chart renders a fixed list of env vars. Anything it doesn't know about — an integration API key, a storage host, a CORS origin — can't be supplied without forking the template. This adds the standard Helm passthrough knobs, every one defaulting to today's render.

| Value | Default | Notes |
|---|---|---|
| `extraEnv` / `extraEnvFrom` | `[]` | appended **after** the chart's own env, so an indexed array entry can be overridden |
| `extraVolumes` / `extraVolumeMounts` | `[]` | appended after the chart's `/tmp` pair |
| `podAnnotations` | `{}` | pod template only — a checksum annotation on the Deployment wouldn't roll pods |
| `initialUser.enabled` | `true` | the InitialUser secretKeyRefs were unconditional and non-optional, so a pod sat in `CreateContainerConfigError` without that secret even with seeding off |
| `auth.secretKey` | `signing-key` | the JWT key name inside `auth.secretName` |
| `tmp.sizeLimit` | `""` | unset keeps `emptyDir: {}` |

### Backward compatibility (proven, not assumed)

`helm template` from `main` and from this branch is **byte-identical** in three cases: default values; a values file exercising only the pre-existing keys the way a real deployment does; and a partial override of `auth:` and `initialUser:` — the two blocks that gained sub-keys — to prove deep-merge still supplies the new defaults. An existing install upgrading with no new values behaves identically.

### Two things documented in `values.yaml` because consumers will hit them

- **Arrays.** `__`-indexed env vars layer over the image's `appsettings.json` arrays rather than replacing them: an array can be extended or overridden by index, never shortened — a removed CORS origin stays live. Authoritative array config belongs in a mounted appsettings file (a follow-up).
- **Seeding.** `Data:SeedSampleData` is the master switch for *all* seeding, initial user included, and with seeding on but no `InitialUser` configured the seed falls back to a well-known default credential. `initialUser.enabled: false` does **not** turn seeding off — only disable it when seeding is off or the database already has users. The values comment says exactly this.

### Deliberately not in this PR

The `endatix-api.env` template helper (it pays off paired with a migration Job, which is separate work) and an Ingress template.

## Test plan

- [x] `helm lint helm` clean (only the pre-existing `icon is recommended` info)
- [x] Byte-identical `helm template` vs `main` for default, deployment-style, and partial-override values
- [x] Positive render with every knob set: annotations on the pod template only, custom secret key, InitialUser refs absent, `sizeLimit` present, `envFrom` present, extra mounts/volumes after the chart's own, `extraEnv` entries after every chart-owned entry with a duplicated name resolving to the later one
- [x] `helm package` at a CI version succeeds; `kubectl apply --dry-run=client` accepts both renders

## Type of Change

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -11,6 +11,10 @@ spec:
       {{- include "endatix-api.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels:
         {{- include "endatix-api.selectorLabels" . | nindent 8 }}
     spec:
@@ -102,7 +106,7 @@ spec:
           valueFrom:
             secretKeyRef:
               name: {{ .Values.auth.secretName }}
-              key: signing-key
+              key: {{ .Values.auth.secretKey }}
         - name: Endatix__Auth__Providers__EndatixJwt__AccessExpiryInMinutes
           value: {{ .Values.auth.accessExpiryInMinutes | quote }}
         - name: Endatix__Auth__Providers__EndatixJwt__RefreshExpiryInDays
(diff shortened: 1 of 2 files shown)


PeterMarinov (approved):
Approved

Issue #899, opened 2026-07-24
Fix negative ExportFormat seed IDs from hashtextextended in InitialReporting migration


PeterMarinov:
### Goal
Ensure Reporting default seed IDs for `ExportFormats` and `SurveyTypeExportMappings` are always positive so they match Snowflake ID conventions and pass Hub `validateEndatixId` / domain `NegativeOrZero` guards.

### Steps to Reproduce
1. Apply `20260721062723_InitialReporting` (PostgreSQL) against a DB with at least one tenant.
2. Inspect `reporting."ExportFormats"` / `reporting."SurveyTypeExportMappings"`.
3. Use a seeded `ExportFormatId` from Hub export (passes through `validateEndatixId`).

### Actual
Seed SQL used raw `hashtextextended(...)`, which returns a signed `bigint`. Roughly half of seeded IDs are negative (e.g. `-6837900068623339226`). Hub rejects them (`must be a numeric string containing only digits` / `must be greater than 0`), so reporting export via Hub fails for those formats.

### Expected
Seeded IDs are deterministic **and** strictly positive.

### Implementation Notes
- Migration has **not** shipped publicly yet (feature-flagged / unpushed). Safe to edit `InitialReporting` in place — no follow-up data repair migration required for public installs.
- Replace each seed id expression:
  - Before: `hashtextextended(t."Id"::text || ':csv', 0)`
  - After: `GREATEST(1, hashtextextended(t."Id"::text || ':csv', 0) & 9223372036854775807::bigint)`
- Apply the same pattern for `:json`, `:codebook`, `:default-map`, and the mapping FK to CSV.
- Do **not** use `abs()` — `abs(long.MinValue)` overflows.
- Local DBs that already ran the old seed: drop/recreate the `reporting` schema (or DB) and re-migrate, or rewrite negative PKs in place (drop FK → update IDs → recreate FK). Do not `INSERT` clones — that hits `IX_ExportFormats_TenantId_Name`.
- File: `oss/src/Endatix.Modules.Reporting/Persistence/Migrations/PostgreSql/20260721062723_InitialReporting.cs`

PR #1016, opened 2026-09-05
fix(#1015): ExportFormat Id collision when seeding default export formats for a new tenant

### LINKED ISSUE #1015: Fix ExportFormat Id collision when seeding default export formats for a new tenant
### Goal

Make `ExportFormatRepository.SeedDefaultsAsync` succeed for a tenant that has no export formats yet. Today it throws before writing anything, so a newly provisioned tenant ends up with no default CSV / JSON / Codebook formats.

### Steps to Reproduce

1. Register the Reporting module in a host.
2. Resolve `IDefaultExportFormatsSeeder` and call `SeedAsync(tenantId, ct)` for a tenant with no rows in `reporting."ExportFormats"`.

### Actual

`AddRangeAsync` throws:

> The instance of entity type 'ExportFormat' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked.

All three new `ExportFormat` instances still have `Id == 0` when they are handed to the change tracker. `BaseEntity.Id` is `[DatabaseGenerated(DatabaseGeneratedOption.None)]`, so `0` is a real key value rather than an EF temporary one — the second and third entity collide with the first.

Ids are assigned in `ReportingDbContext.ProcessEntities()`, which runs from `SaveChanges` / `SaveChangesAsync`. That is after tracking, so it never gets the chance to run for this call.

### Expected

The three default formats and the default mapping are created, each with a distinct Snowflake Id.

### Implementation Notes

Assign Ids with an EF value generator, so they exist by the time an entity is tracked rather than at `SaveChanges`. `ReportingDbContext` takes the DI-registered `EfCoreValueGeneratorFactory` and wires a generator for every `BaseEntity` in its model:

```csharp
builder.Entity(entityType.ClrType)
    .Property<long>(nameof(BaseEntity.Id))
    .HasValueGenerator((property, _) => valueGeneratorFactory.Create<long>(property))
    .ValueGeneratedNever();
```

This is the pattern `AppDbContext`, `AppIdentityDbContext` and `JobsDbContextBase` (#959) already use. It covers every multi-row `Add` in the module rather than one call site, and leaves the module with one Id mechanism instead of two.

**Inject the factory; do not construct one per context.** EF caches one model per context type, so the factory belonging to the first context built in a process is captured for that model's lifetime. Two test doubles return unusable Ids and will silently defeat the generator for every later context in the assembly — `ReportingTestSchema` returns `0`, and `ReportingDbContextTests` substitutes a fixed pair. Both need a real incrementing generator, or the fix appears not to work.

Worth fixing in the same change, all reachable through the tenant-provisioning path this method exists for:

- The idempotence guards run under the tenant query filter, so seeding tenant A from tenant B's scope reads `TenantId == B AND TenantId == A` and matches nothing. Use `IgnoreQueryFilters` with explicit `TenantId`/`IsDeleted` predicates, the shape `FlattenedSubmissionRepository` already uses. The duplicate rows cannot actually land — the unique index on `(TenantId, Name)` is filtered on `IsDeleted` and is not tenant-scoped — so the symptom is a wasted insert failing with `23505` as an unhandled `DbUpdateException`.
- Formats and mapping commit in two `SaveChanges` with no transaction. A crash between them leaves formats present, so the guard returns early and the default mapping can never be created on a retry.
- A concurrent provisioning race throws an opaque 500.
- The mapping guard skips whenever a mapping exists, so once the formats are soft deleted it keeps pointing at a deleted one — and `GetTenantDefaultAsync` resolves it through an `Include` that the soft-delete filter nulls out, leaving the tenant with no default forever.

Context:

- Line 223 is currently the only EF `AddRange` in the Reporting module, which is why single-entity paths have never surfaced this. The default `SurveyTypeExportMapping` added later in the same method uses `AddAsync`, so it is unaffected.
- `IDefaultExportFormatsSeeder` is registered in `ReportingModule` but has no production caller today; the failure is currently only reachable from test setup. Wiring the seeder into tenant provisioning — what its doc comment describes — makes this break new-tenant creation, so it is worth fixing before that lands.
- Related: #899 covered the same table's **migration** seed path. This is the **runtime** seed path used for tenants provisioned after that migration has run.
- The default formats are defined twice, once in the `InitialReporting` migration SQL and once as constants in the repository, and already disagree on `locale`. Pre-existing and orthogonal to this fix.

Files:

- `src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs` (`SeedDefaultsAsync`)
- `src/Endatix.Modules.Reporting/Persistence/ReportingDbContext.cs` (constructor + `OnModelCreating`)
- `src/Endatix.Infrastructure/Data/AppDbContext.cs` (`ConfigureEntityIdValueGenerators`, the pattern to mirror)
- `tests/Endatix.IntegrationTests/Infrastructure/ReportingTestSchema.cs` and `tests/Endatix.Modules.Reporting.Tests/Persistence/ReportingDbContextTests.cs` (the id generators that must not return 0 or a fixed value)

igdragoev:
## Description

`ExportFormatRepository.SeedDefaultsAsync` threw before writing anything. It handed three `ExportFormat` rows to `AddRangeAsync` while their Ids were still `0`, and since `BaseEntity.Id` is `DatabaseGeneratedOption.None`, `0` is a real key value rather than an EF temporary one — so the change tracker rejected the second row:

> The instance of entity type 'ExportFormat' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked.

Ids were stamped only by `ReportingDbContext.ProcessEntities`, which runs from `SaveChanges` — after tracking, so it never got the chance.

### Ids are now assigned when an entity is tracked

`ReportingDbContext` takes the DI-registered `EfCoreValueGeneratorFactory` and wires a value generator for every `BaseEntity` in its model, through a shared `ConfigureEntityIdValueGenerators` extension. This is the pattern `AppDbContext`, `AppIdentityDbContext` and `JobsDbContextBase` (#959) already use. Every multi-row `Add` in the module is covered, not just this call site, and the module is left with one Id mechanism instead of two.

**Injecting the factory rather than constructing one per context is load-bearing.** EF caches one model per context type, so the factory belonging to the first context built in a process is captured for that model's lifetime. Two test doubles returned unusable Ids and would have silently defeated the generator for every later context in the assembly — `ReportingTestSchema`'s generator returned `0`, and `ReportingDbContextTests` substituted a fixed pair. Both now increment, with the reason recorded where they are defined.

### Seeding correctness

Four further problems, all reachable through the tenant-provisioning path this method exists for:

| Problem | Fix |
|---|---|
| The idempotence guards ran under the tenant query filter, so seeding tenant A from tenant B's scope read `TenantId == B AND TenantId == A` and matched nothing | `IgnoreQueryFilters` with explicit `TenantId`/`IsDeleted` predicates — the shape `FlattenedSubmissionRepository` already uses |
| Formats and mapping committed in two `SaveChanges` with no transaction; a crash between them left the mapping uncreatable on any retry, because the formats-exist guard then returned early | One transaction, one save |
| A concurrent provisioning race threw an opaque 500 | The loser re-checks and returns quietly when the winner produced the state it wanted |
| The mapping guard skipped whenever a mapping existed, so once the formats were soft deleted it kept pointing at a deleted one — and `GetTenantDefaultAsync` resolves it through an `Include` that the soft-delete filter nulls out, so the tenant read as having no default forever | The mapping is repointed at a live format; adds `SurveyTypeExportMapping.UpdateExportFormat` |

On the first row: the duplicate rows could never actually land, because the unique index on `(TenantId, Name)` is filtered on `IsDeleted` and is not tenant-scoped. What the defeated guards produced was a wasted insert failing with `23505`, surfacing as an unhandled `DbUpdateException`.

### Deliberately out of scope

- `AppDbContext` and `JobsDbContextBase` still carry private copies of the same wiring. Deduplicating onto the shared extension is worth doing once #959 lands, rather than conflicting with it now.
- The default formats are defined twice — once in the `InitialReporting` migration SQL, once as constants here — and already disagree on `locale`. Pre-existing and orthogonal.
- `ReportingDbContext` keeps its `ProcessEntities` Id branch as a no-op fallback. #959 restructures that method, so leaving it alone keeps the conflict surface small.

## Related Issues

Fixes #1015. Related to #899 (the *migration* seed path for the same table) and #959, which introduces the value-generator pattern this follows and touches the same `ReportingDbContext` methods.

## Type of Change

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

- I have added tests that prove my fix is effective

> [!NOTE]

## Testing

Seven PostgreSQL integration tests: the multi-row insert, per-format settings, idempotence, the cross-tenant scope, soft-delete repair, and a missing mapping. Plus a model-level fan-out test over **both** providers, which needs no database and covers the SQL Server leg the integration tests skip.

Each was checked to fail without the code it guards:

- Remove the generator wiring → all 7 integration tests and both provider legs fail.
- Remove the guards and the race handling → the cross-tenant test fails with `23505 duplicate key value violates unique constraint "IX_ExportFormats_TenantId_Name"`.

Suites: `Endatix.IntegrationTests` 108 passed / 1 skipped · `Endatix.Modules.Reporting.Tests` 396 passed / 1 skipped · `Endatix.Infrastructure.Tests` 1183 passed.

### DIFF
--- a/src/Endatix.Infrastructure/Data/DbContextModelBuilderExtensions.cs
+++ b/src/Endatix.Infrastructure/Data/DbContextModelBuilderExtensions.cs
@@ -78,6 +78,47 @@ public static class DbContextModelBuilderExtensions
     }
 
     /// <summary>
+    /// Configures Snowflake Id generation for every <see cref="BaseEntity"/> in the model, so an Id is
+    /// assigned when the entity starts being tracked rather than at SaveChanges time.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <see cref="BaseEntity.Id"/> is <c>DatabaseGeneratedOption.None</c>, so the default value of 0 is a
+    /// real key value rather than an EF temporary one. Without an Add-time generator, adding more than
+    /// one new entity before saving puts several rows with Id 0 into the identity map and the change
+    /// tracker rejects the second as a duplicate key.
+    /// </para>
+    /// <para>
+    /// EF caches one model per context type, so the <paramref name="valueGeneratorFactory"/> passed by
+    /// the first context built in a process is captured for the lifetime of that model. Pass the
+    /// DI-registered singleton rather than constructing one per context, or every later context will
+    /// silently generate ids from the first one's generator.
+    /// </para>
+    /// </remarks>
+    /// <param name="builder">The model builder to configure.</param>
+    /// <param name="valueGeneratorFactory">Factory producing the Snowflake-backed value generator.</param>
+    public static void ConfigureEntityIdValueGenerators(
+        this ModelBuilder builder,
+        EfCoreValueGeneratorFactory valueGeneratorFactory)
+    {
+        Guard.Against.Null(builder, nameof(builder));
+        Guard.Against.Null(valueGeneratorFactory, nameof(valueGeneratorFactory));
+
+        var entityTypes = builder.Model.GetEntityTypes()
+            .Where(entityType =>
+                !entityType.IsOwned() &&
+                typeof(BaseEntity).IsAssignableFrom(entityType.ClrType));
+
+        foreach (var entityType in entityTypes)
+        {
+            builder.Entity(entityType.ClrType)
+                .Property<long>(nameof(BaseEntity.Id))
+                .HasValueGenerator((property, _) => valueGeneratorFactory.Create<long>(property))
+                .ValueGeneratedNever();
+        }
+    }
+
+    /// <summary>
     /// Applies entity type configurations from the specified assembly, filtered by DbContext type using generic attributes.
     /// This allows isolating configurations for different DbContexts that share the same assembly.
     /// </summary>
--- a/src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
+++ b/src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
@@ -1,4 +1,5 @@
 using System.Text.Json;
+using Endatix.Core.Abstractions;
 using Endatix.Modules.Reporting.Contracts.Export;
 using Endatix.Modules.Reporting.Domain;
 using Endatix.Modules.Reporting.Features.Export;
@@ -14,7 +15,8 @@ internal sealed class ExportFormatRepository(
     ReportingDbContext dbContext,
     IReportingUnitOfWork unitOfWork,
     ExportFormatSettingsParser settingsParser,
-    IExportCapabilityRegistry capabilityRegistry) : IExportFormatRepository
+    IExportCapabilityRegistry capabilityRegistry,
+    IIdGenerator<long> idGenerator) : IExportFormatRepository
 {
     private static readonly string _defaultSubmissionsSettingsJson = JsonSerializer.Serialize(new
     {
@@ -220,7 +222,18 @@ internal sealed class ExportFormatRepository(
             "Default form definition codebook export");
         codebookFormat.UpdateSettingsJson(_defaultCodebookSettingsJson);
 
-        await dbContext.ExportFormats.AddRangeAsync([csvFormat, jsonFormat, codebookFormat], cancellationToken);
+        ExportFormat[] defaultFormats = [csvFormat, jsonFormat, codebookFormat];
+
+        // Ids must be assigned before the rows are tracked. BaseEntity.Id is DatabaseGeneratedOption.None,
+        // so the default 0 is a real key value rather than an EF temporary one, and the change tracker
+        // rejects the second row of the batch as a duplicate key. The SaveChanges-time stamping in
+        // ReportingDbContext.ProcessEntities only covers rows added one at a time.
+        foreach (var format in defaultFormats)
+        {
+            format.Id = idGenerator.CreateId();
+        }
+
+        await dbContext.ExportFormats.AddRangeAsync(defaultFormats, cancellationToken);
         await unitOfWork.SaveChangesAsync(cancellationToken);
 
         var hasDefaultMapping = await dbContext.SurveyTypeExportMappings
--- a/src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
+++ b/src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
@@ -1,5 +1,4 @@
 using System.Text.Json;
-using Endatix.Core.Abstractions;
 using Endatix.Modules.Reporting.Contracts.Export;
 using Endatix.Modules.Reporting.Domain;
 using Endatix.Modules.Reporting.Features.Export;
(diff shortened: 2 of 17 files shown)


PeterMarinov:
@igdragoev - thanks for the write-up and for isolating [#1015](https://github.com/endatix/endatix/issues/1015). After reviewing, I have reached the conclusion that it will be best if we close this PR in favour of a smaller fix on the XLSX stack, which was already opened before this PR. Let me share the thinking process and argumentation:

## What we understood

This PR addressed an issue that `SeedDefaultsAsync` failed before it wrote anything: several `ExportFormat` rows were tracked with `Id == 0` while `BaseEntity.Id` is `DatabaseGeneratedOption.None`, so EF treated `0` as a real key. Ids are stamped in `ReportingDbContext.ProcessEntities` on `SaveChanges`, which is **after** tracking — so `AddRange` never got that far.

You also called out follow-on issues on the same path:

- tenant query filters hiding the target tenant’s rows
- formats vs mapping in two saves
- concurrent provision races
- default mapping stuck on a soft-deleted format

The **Id collision is real**. The extra work around value generators, transactions, unique-violation swallow, and mapping repair is more than we want to take for this bug, and it collides with [#959](https://github.com/endatix/endatix/issues/959).

## How we cover the same bug

Two small PRs on `feature/e1012-xlsx-export` (Excel export + tenant seed):

1. **[#1014](https://github.com/endatix/endatix/pull/1014)** — `SeedDefaultsAsync` inserts **one row per `SaveChanges`**. Tracker never holds two unsaved `Id = 0` entities. Same outcome as stamping snowflakes before `AddRange` (your issue notes / `DataSeeder`), without changing how Reporting assigns Ids.

2. **[#1026](https://github.com/endatix/endatix/pull/1026)** — seed lookups use `IgnoreQueryFilters()` + explicit `TenantId` / `!IsDeleted`. Required because `tenant.created` is app-level: `Tenant` is not `ITenantOwned`, so the outbox row is `TenantId = 0` while the payload has the real id. A filtered read would look empty and try to insert again.

Together that is the production path we are wiring: outbox `tenant.created` → `IDefaultExportFormatsSeeder`. Existing tenants get Excel from the PG data migration on #1014, not from this PR.

Also, I have created a new issue to establish better pattern for the Ids so they are set before SaveChanges and can be used for integration and in process events same as Jobs on Add/AddRange - https://github.com/endatix/endatix/issues/1027, which we will plan to do very soon and address this friction

## What we are not taking from this PR (and why)

| This PR | Why we skip it |
|---|---|
| `ConfigureEntityIdValueGenerators` on `ReportingDbContext` | Fixes more than seed; fights #959; EF still `ValueGeneratedNever()`, so generators are easy to misread as protection they do not give. Leave Id wiring to #959. Added https://github.com/endatix/endatix/issues/1027 to establish better pattern |
| One transaction + unique-violation “loser retries” | First-time provision + outbox retry. Four default rows. Not worth the opacity yet. |
| Soft-delete remapping / `UpdateExportFormat` | Real edge case, not #1015. Open a follow-up if we see stuck defaults in prod. |
| Locale / duplicate default definitions | Orthogonal; #1014 already aligned XLSX migration JSON with runtime settings. |

We will not merge a 600-line Reporting DbContext change to unblock seeding four rows.

## Ask

Please look at #1026 (and the seed loop on #1014). If you still want value-generator unification, keep that on #959 rather than this seed bug.

Closing as **not merged** — #1015 is addressed by #1014 + #1026.

Issue #862, opened 2026-07-13
epic: Add async background job infrastructure


PeterMarinov:
### Goal

Introduce a generic **Background Jobs** infrastructure for long-running tenant operations (submission export first, then GDPR purge, backfill, schema recompile). API returns `HTTP 202` + `jobId`; Hub polls job status until complete.

### Context

- Sync streaming export is shipped (E6/E7 reporting read model).
- Large exports need decoupled processing, progress tracking, and resumable download — not a bespoke export-only polling stack.
- Aligns with existing **transactional outbox** + in-process relay (`AddEndatixOutboxRelay`); no direct API→worker coupling.

### Scope (iterative)

1. `BackgroundJob` entity + single-row enqueue; the job table is the queue
2. Generic `GET /jobs/{jobId}` status API (tenant-scoped)
3. `IBackgroundJobHandler` registry; first handler: **SubmissionExport**
4. Hub async export UX (poll → download)

### Out of scope (follow-up)

- External message bus (Dapr/RabbitMQ) — in-process relay first
- Auto-threshold sync→async routing
- Email notifications on job complete

### Plan

See `.cursor/plans/e*-background-jobs-plan.md` in endatix-saas (linked from export epic #797).

### Related

- Export epic: #797
- E7 export settings: #861

igdragoev:
## Design note for PR-J3: job runtime vs. the outbox relay

The plan currently has the outbox subscriber run the whole job inline ("Handler loads job, sets `Processing`, runs work, updates progress, sets `Completed`/`Failed`"). That collides with the relay's mechanics in two ways for any job longer than `Endatix:Outbox:LeaseSeconds` (default **60s**) — which is the normal case for large exports.

### Problem 1: head-of-line blocking

The relay is a single sequential loop (claim batch → `await PublishAsync` per message → next tick). A job running inline in `PublishAsync` stalls the loop: every other outbox message, **including all webhook deliveries**, waits behind it for the job's full duration. On a single-instance deployment (the OSS default), one large export freezes the entire outbox pipeline — a user clicking Export would delay every webhook by minutes. The rest of the claimed batch (up to `BatchSize` rows) also sits leased and unprocessed while the loop is stuck.

### Problem 2: lease expiry → duplicate execution (multi-instance)

1. `t0` — instance A claims the `JobRequestedEvent` row (lease `t0+60s`) and starts the export.
2. `t+60s` — the lease expires while A is still working. The row is still `Pending`, so it's claimable again.
[...]

PR #1021, opened 2026-09-08
feat(e862): add the background job status endpoint

### LINKED ISSUE #1028: Add GET job status endpoint
### Goal

Ship `GET /api/jobs/{jobId}` so a caller that started long-running work can poll tenant-scoped job state without holding the original request open (e862 slice J2).

### Implementation Notes

- Parent: #862
- PR: https://github.com/endatix/endatix/pull/1021
- Status-only contract (id, type, status, progress, status/error messages). Do not publish `ResultJson` from this endpoint.
- Tenant isolation must be explicit (`TenantId <= 0` → unauthorized; filter on `TenantId` in the query). Ambient EF tenant filter treats `0` as bypass.
- Permission model still open: no `Permissions(...)` yet; settle before treating the slice as done (plan §2.6 is job-type-dependent).

igdragoev:
## Description of changes

`GET jobs/{jobId}` — e862 slice J2, on FastEndpoints. A caller that started long-running work can follow it without holding a request open.

| | |
|---|---|
| Endpoint | `Endpoints/Jobs/GetJob.cs`, `GetJob.JobResponse.cs` |
| Handler | `Features/GetJob/GetJobHandler.cs` — query, DTO, and the read |
| Module | `JobsModule` implements `IHasFastEndpoints`, so its endpoints are discovered only when the module registers |
| Core | `JobStatus` gains `[JsonConverter(typeof(JsonStringEnumConverter))]`, matching how the platform's other enums cross the wire |

```json
{"id":"987654321098765432","type":"SubmissionExport","status":"Processing",
 "progressPercentage":45,"statusMessage":"Processing 4,500 of 10,000 rows",
 "errorMessage":null}
```

Identifiers serialize as strings through the existing `LongToStringConverter`, so a snowflake survives a JavaScript client intact.

## Related
- closes https://github.com/endatix/endatix/issues/1028

## Tenant scoping

**The ambient query filter cannot carry this on its own**, and an earlier revision of this PR relied on it. That filter reads a tenant id of zero as *no tenant, show everything*, so background services can sweep across tenants — and `TenantContext.TenantId` is a plain `long` defaulting to zero, which `TenantMiddleware` leaves unset on three early-return paths. An authenticated principal with no parseable `tid` claim still satisfies the default policy, so the endpoint would have served it any tenant's job.

The handler now refuses a request with no tenant and filters on `TenantId` directly, matching `GetExportFormatHandler` and `GetFolderBySlugHandler`. Verified against a running host: a token identical to a working one but with the `tid` claim removed gets `401 "Tenant context is required."`.

A job belonging to another tenant is **not found** rather than forbidden — a 403 would confirm the id exists, which is enough to enumerate what another tenant is running.

## Permission

`Actions.Jobs.View` — `jobs.view`, following the documented `{category}.{action}` pattern, in a category of its own. Borrowing `submissions.export` was the alternative and would have tied a generic status endpoint to one job type, making it the wrong gate for webhook delivery.

**Granted to Creator only.** Admin and PlatformAdmin satisfy every permission check without a grant, because `AuthorizationDataExtensions.HasPermission` short-circuits on `IsAdmin` — which is why Admin's seeded list is just `access.apps.hub`. Seeded in both providers, since `AppIdentityDbContext` is dual-provider; both inserts are conditional and `Down` removes them.

Deliberately coarse. Reporting on a job says only how far the work got, and the work itself is reached through the endpoint that owns it, carrying its own permission. Plan §2.6's per-job-type rules belong with the job types that need them.

## The response carries nothing about what the job produced

An untyped result is not a contract. One queue serves every job type, so the only shape the response could promise across all of them is a free-form object, which a client cannot read without switching on `type` and relying on knowledge the contract does not carry.

Where a job produced something, the endpoint that owns that output exposes it — an export's file name and content type belong on the download response as headers, not restated in a status body. `ResultJson` stays on the job row for the handler, the download path and operators; it is simply not published. Re-adding is additive if a job type ever has output a caller genuinely must read, and the right form then is a shape keyed on job type.

## Errors use the platform envelope, with no wiring of its own

Every failure path returns the canonical RFC 7807 body as `application/problem+json`, because `Results<Ok<T>, ProblemHttpResult>` routes through `ToProblem` and FastEndpoints validation already goes through the `Errors.ResponseBuilder` set up in endatix#1001. Verified against a running host:

**Validation** — `GET /api/jobs/0`:

```json
{"type":"...#name-400-bad-request","title":"There was a problem with your request","status":400,
 "detail":"'job Id' must be greater than '0'.","instance":"/api/jobs/0","traceId":"00-f27c9f...",
 "errorCode":"GreaterThanValidator","fields":{"jobId":["'job Id' must be greater than '0'."]}}
```

**Binding** — `GET /api/jobs/not-a-number` produces the envelope with `fields` and no `errorCode`, since a bind failure carries no validator name. **Not found** returns `detail` and `traceId` with no `fields`, and is what a cross-tenant request gets too. All four statuses — 200, 400, 401, 404 — are declared on the operation and appear in the generated document.

Worth contrasting with the spike in #1020, where the same responses each needed explicit registration. On this path none of that exists: the endpoint inherits the conventions by being a FastEndpoints endpoint.

## Verification

Unit: 45 in `Endatix.Modules.Jobs.Tests`, 9 of them new — endpoint mapping, validator, the tenant guard, and the handler against a real in-memory context so the tenant filter is exercised rather than assumed.

The permission was verified against a live host with the migration applied: a token carrying `jobs.view` gets 200, one without it gets 403, and the seeded row and Creator grant are present in `identity.Permissions` / `identity.RolePermissions`.

Also run against a live host and PostgreSQL, covering the status shape, cross-tenant and no-tenant refusals, the validation and binding 400s, and the generated OpenAPI document.

Core 1432, Api 892, Hosting 146 all pass.

Refs #862


igdragoev (commented):
### Review — `GET /api/jobs/{jobId}`

Six inline groups below, ordered by how much they matter. Two PR-level notes here.

---

**Tenant isolation is verified only against the EF InMemory provider.**
`tests/.../Features/GetJob/GetJobHandlerTests.cs:79`

The api rules ask for integration tests covering "endpoint wiring, authentication/authorization, DI registration, middleware, EF mappings, and real database behavior", Testcontainers-backed. This PR ships a new route whose entire authorization design is the global query filter, covered by a single InMemory unit test. Nothing asserts a 401, nothing asserts the route is served once `JobsModule` registers, and nothing covers a caller whose `TenantId` resolves to 0 — which is the case that actually fails (see the first inline comment). The cross-tenant 404 and 401 evidence in the PR body is a manual run against a live host, so nothing guards those in CI.

**FastEndpoints and FluentValidation are used directly but arrive transitively.**
`src/Endatix.Modules.Jobs/Endatix.Modules.Jobs.csproj:34`

`GetJob.cs` has `using FastEndpoints;` and `using FluentValidation;`, but the csproj adds only the `Endatix.Api` ProjectReference. `Endatix.Modules.Reporting.csproj` — the reference module doing the same thing — declares `<PackageReference Include="FastEndpoints" />` explicitly.

Worth naming at altitude too: `ARCHITECTURE.md` documents the module flow as `Modules.* -> Core, Infrastructure, Modules.*.Contracts`. Referencing the whole of `Endatix.Api` pulls in every endpoint, Swagger and `FastEndpoints.Security` for one 37-line `IHasFastEndpoints` interface, and permanently forecloses any future `Endatix.Api -> Endatix.Modules.Jobs` reference.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+/// No tenant argument: the jobs context filters every query to the current tenant, so a job
+/// belonging to another one is simply not found. Passing a tenant id here as well would add a
+/// second, weaker guard that could disagree with the first.
+/// </remarks>
+internal sealed class GetJobHandler(IJobsDbContext dbContext)
+    : IQueryHandler<GetJobQuery, Result<JobDto>>
+{
+    public async Task<Result<JobDto>> Handle(GetJobQuery request, CancellationToken cancellationToken)
+    {
+        var job = await dbContext.BackgroundJobs
```
igdragoev, line 54: **Tenant isolation can fail open when the ambient tenant resolves to 0.**

The only thing scoping this read to a tenant is the global query filter, which is `GetTenantId() == 0 || e.TenantId == GetTenantId()` (`DbContextModelBuilderExtensions.cs:47-58`). `JobsDbContextBase.GetTenantId()` returns 0 outside a request deliberately, so the sweeper can scan every tenant — but `TenantContext.TenantId` also stays 0 *during* a request whenever the principal carries no `tid` claim:

- `TenantMiddleware` returns early if the claim is missing or unparseable;
- `ClaimsTransformer.GetAuthorizationDataAsync` returns null — so no `AuthorizedIdentity` and no `tid` — when the user id is missing, when no `IAuthorizationStrategy` matches the issuer, or when any exception is thrown (it is caught and logged).

Such a principal still satisfies the default policy (authenticated + `sub`), so `GET /api/jobs/{id}` returns another tenant's job row.

Related, and part of the same fix — **`GetJob.cs:26` declares no permissions.** The api rules require a `Permissions(...)` call immediately after the verb. `GetJob` is one of only four endpoints in the product with no `Permissions`/`Roles`/`Policies`/`AllowAnonymous`; the other three (`Auth/Logout`, `Auth/Me`, `MyAccount/ChangePassword`) are self-scoped by construction. With none, the fallback is the default policy alone — which the tenant-0 principal above passes.

Every other tenant-scoped read guards this explicitly: `GetFolderBySlugHandler` opens with `Guard.Against.NegativeOrZero(tenantContext.TenantId)`, and Reporting's `Settings/ExportFormats/Get.cs` passes `tenantContext.TenantId` into the query. The remark below calls that explicit check "a second, weaker guard" — it is the stronger one, and the only thing that closes the tenant-0 hole.

Suggest both: `Guard.Against.NegativeOrZero(tenantContext.TenantId)` here, and a `Permissions(...)` declaration on the endpoint.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.JobResponse.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// What the job produced, in the shape its handler wrote. Null until the job completes, and for
+    /// job types that produce nothing.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately untyped. One queue serves every job type, and they produce different things — an
+    /// export produces a file, a webhook delivery produces a response code — so there is no single
+    /// shape this endpoint could promise. Each job type documents its own, and a caller already knows
+    /// which type it asked about from <see cref="Type"/>.
+    /// </remarks>
+    public JsonNode? Result { get; init; }
```
igdragoev: **The untyped `result` passthrough has four consequences worth settling together.**

The design intent is clear from the remark on `ParseResult` and I don't disagree with it — a job type's output shape is the job type's business. But `JsonNode?` is the wrong vehicle:

**1. It removes the only allow-list on what crosses the wire.** (`GetJobHandler.cs:95`) `ResultJson` is written by whichever handler ran the job and is mapped `jsonb` with no length cap (`BackgroundJobConfigurationPostgreSql.cs:26-28`) — unlike `StatusMessage`/`ErrorMessage`, capped at 512/2048. The old `JobResultDto(DownloadUrl, FileName, ContentType)` meant three named fields could reach a caller; now the whole column does, on an endpoint with no permission declaration. A future export handler that stores an internal storage key, a pre-signed URL or an upstream response body next to the file name publishes all of it. A large payload is also fully materialized into a node tree and re-serialized on every poll.

**2. It generates a broken OpenAPI schema.** `JsonNode` is an abstract class whose public surface is `Parent` (`JsonNode?`), `Root` (`JsonNode`) and `Options`. NSwag reflects over unrecognised types, so `Produces<JobResponse>(200)` emits a `result` schema of `parent`/`root`/`options` with a self-reference rather than a free-form object — and `api-contracts` generates the Kotlin, .NET and Orval clients from that document.

**3. The repo already has the right type for this.** `JsonElement?` is what `FormModel.cs:89` (WebHookSettings) and `Create.CreateFormRequest.cs:49,60` use for exactly this situation, and it serializes to a free-form object.

**4. `ParseResult` duplicates an existing helper.** (`GetJobHandler.cs:86`) `src/Endatix.Api/Endpoints/Forms/FormMapper.cs:40` is the same function line for line — null/whitespace guard, parse, `catch (JsonException) => null` — and already returns `JsonElement?`. It is `internal`, so at minimum the convention is at hand.

Switching this property to `JsonElement?` resolves 2, 3 and 4 at once and keeps the design intent intact. 1 is a separate call about whether an uncapped, handler-written column should reach clients verbatim.

Also: **`GetJob.cs:32` is now stale.** The Swagger description still reads "Download metadata appears only once the job has completed successfully", after `JobResultResponse { downloadUrl, fileName, contentType }` was removed. A client reading the generated document is told to expect a `downloadUrl` that is not in the schema.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+            job.Id,
+            job.JobType,
+            job.Status,
+            job.ProgressPercentage,
+            job.StatusMessage,
+            // Only a completed job has produced anything. A failed one may still carry the result of
+            // an earlier successful attempt, and offering that as this job's output would hand the
+            // caller a stale artifact under a job that did not finish.
+            job.Status == JobStatus.Completed ? ParseResult(job.ResultJson) : null,
+            job.ErrorMessage));
```
igdragoev: **`ErrorMessage` is operator-facing text and is returned verbatim in a 200 body.**

`BackgroundJobConfiguration.cs` documents the column as "an exception message for an operator to read... The full detail belongs in logs, correlated by TraceId." `AGENTS.md` is firmer: `ex.Message` from EF Core/Npgsql/BCL "carries connection strings, SQL and file paths" and must never reach the caller.

`ResultFactoryMustNotInterpolateExceptionMessageTests` is the guard for that, but its regex only matches `Result.Error|Invalid|NotFound|...(` call sites — a runner writing `job.Fail(ex.Message, now)` is invisible to it. The runner does not exist yet, so nothing currently constrains what lands in the column, and this endpoint is what makes it reachable by a client.

Two ways to close it, either is fine: sanitize at the endpoint, or make "must be author-written, end-user-safe text" an enforced contract on `BackgroundJob.Fail`/`DeadLetter` before the read ships.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
```diff
@@ -0,0 +1,76 @@
...
+            summary.Summary = "Get a background job";
+            summary.Description =
+                "Retrieves the current state of a background job. Progress and status message are " +
+                "advisory and may lag the work itself; status is authoritative. Download metadata " +
+                "appears only once the job has completed successfully.";
+            summary.ExampleRequest = new GetJobRequest { JobId = 987654321 };
+            summary.Responses[200] = "Job state retrieved.";
+            summary.Responses[404] = "Job not found.";
+        });
+        Description(builder => builder
```
igdragoev, line 44: **The 400 the validator produces is not declared.**

`GetJobValidator` rejects `JobId <= 0` and FastEndpoints answers 400 problem+json, but neither `summary.Responses[400]` nor `.ProducesProblem(400)` is present.

`AGENTS.md`: "OpenAPI: `Description(b => b.Produces<T>(...).ProducesProblem(400).ProducesProblem(404))` listing exactly the statuses the endpoint's `Summary(s => s.Responses[...])` declares — every endpoint returning `ProblemHttpResult` must have one." The api rules add "Always document errors with ProducesProblem." `GetFormSchemaLocales.cs` declares both.

Generated clients get no 400 branch as it stands.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+        if (string.IsNullOrWhiteSpace(resultJson))
+        {
+            return null;
+        }
+
+        try
+        {
+            return JsonNode.Parse(resultJson);
+        }
+        catch (JsonException)
```
igdragoev: **A malformed result payload is swallowed with no log.**

The handler takes no `ILogger`, so a job handler that writes invalid JSON to `ResultJson` produces a Completed job reporting `result: null` on every poll, forever, with nothing recorded anywhere to explain it. The caller sees success with no output and no error — the hardest shape to diagnose from the outside.

Reporting the read as no-result rather than failing it is the right call; it is the silence that is the problem. Every comparable site in the repo logs: `ExportFormatSettingsParser.cs:30`, `GetFormDefinitionFieldsHandler.cs:83`, `JsonExtractor.cs:28`, `SubmissionExporterBase.cs:137`.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                candidate.Id,
+                candidate.JobType,
+                candidate.Status,
+                candidate.ProgressPercentage,
+                candidate.StatusMessage,
+                candidate.ResultJson,
+                candidate.ErrorMessage,
+            })
+            .SingleOrDefaultAsync(cancellationToken);
```
igdragoev: **Four small cleanups, grouped.**

**`SingleOrDefaultAsync` on a primary-key lookup** (this line). `Id` is the PK, so a second row is impossible — `SingleOrDefaultAsync` fetches two rows to prove it and adds an `InvalidOperationException` branch that can never fire. `FirstOrDefaultAsync` is the correct call. Relatedly, `.AsNoTracking()` on line 44 is a no-op: a projection to an anonymous type is never tracked.

**The comment on line 69 describes a state the domain cannot produce.** "A failed one may still carry the result of an earlier successful attempt" — but `BackgroundJob.Complete` is the only writer of `ResultJson`, and `EnsureStatus` makes Completed terminal (`Fail`, `Reschedule` and `DeadLetter` all require Processing). No non-Completed row can carry a result. The gate itself is harmless defence; the comment sends the next reader looking for a state machine that does not exist, and invites someone to "fix" the guard on a false premise.

**`GetJob.JobResponse.cs:52` contradicts the domain.** "Why the job failed. Set only once it has failed or been dead-lettered." — but `BackgroundJob.Reschedule(nextAttemptAt, errorMessage)` sets `ErrorMessage` and leaves `Status = Retrying`. A client that follows the documented contract and treats a non-null `errorMessage` as terminal will report a job that is merely between attempts as failed, and stop polling it.

**`GetJobEndpointTests.cs:61` asserts less than it claims.** `ExecuteAsync_UnfinishedJob_ReportsNoResult` arranges a `JobDto` whose `Result` is already null and asserts the response `Result` is null — it asserts that `FromDto` copies null. Its comment says "a caller that polls must not be offered an output before there is one", but the gate that enforces that is `job.Status == JobStatus.Completed ? ... : null` on line 72 here: delete that line and the test still passes. It also duplicates `ExecuteAsync_ExistingJob_ReturnsItsState`'s arrangement exactly, differing only in which field it asserts.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
igdragoev: You are right, and the reasoning in my PR description was wrong. I claimed an explicit check "could only agree with the filter or be wrong"; it is the stronger guard, for exactly the reason you give.

Verified before changing anything: `TenantContext.TenantId` is a plain `long` defaulting to 0, and `TenantMiddleware` has three early returns that leave it unset. So an authenticated principal with no parseable `tid` keeps tenant 0, which the filter reads as "no tenant, show everything". That is a cross-tenant read, not a theoretical one.

The handler now refuses a request with no tenant (`Result.Unauthorized`) and filters on `TenantId` directly, matching `GetExportFormatHandler` and `GetFolderBySlugHandler`. Proved against a running host: a token identical to the working one but with the `tid` claim removed now gets `401 "Tenant context is required."` where it previously would have been served the row. A unit test covers it.

On `Permissions(...)`: **not done, and I would rather you decided it than have me invent it.** No jobs permission exists, so adding one means choosing a constant, adding a catalogue entry and deciding which roles are granted it — get that wrong and either legitimate users are locked out or it is granted too broadly. Plan §2.6 makes the answer job-type-dependent ("creator or `submissions.export` on the related form"; tenant admin for system-enqueued jobs), and no job type exists yet to key off. Your count is right that this would be the fourth endpoint without a declaration and the only one that is not self-scoped — the tenant guard closes the hole either way, but the declaration is still owed.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.JobResponse.cs
igdragoev: Switched to `JsonElement?`, which resolves 2, 3 and 4 as you said.

Verified 2 rather than taking it on trust — the generated document now carries `"result": {"description": "...", "nullable": true}`, a free-form object, with no `JsonNode` type anywhere in the schema set. `ParseJsonString` in `FormMapper` is `internal` to `Endatix.Api` so I could not call it, but the implementation now matches it line for line.

The stale Swagger description is fixed too; it was mine, left behind when the shape changed.

On 1 — the uncapped column reaching clients verbatim — I have left the passthrough, since the untyped result was a deliberate decision after a webhook delivery could only have been described as three nulls under the old shape. But I think you are right that it wants a bound rather than trust, and the natural place is the same contract I added for `ErrorMessage`: state on the entity what a handler may put in `ResultJson`, and enforce a length cap, before the first handler writes one in J4. Raising it as an open item rather than closing it here.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
igdragoev: Taken the second option — the contract now lives on the entity rather than being sanitized at the read.

`BackgroundJob.ErrorMessage` documents it directly: must be author-written text safe to show an end user, because the status endpoint returns the column verbatim and exception text from EF Core, Npgsql and the BCL carries connection strings, SQL and file paths; log the exception and correlate by `TraceId` instead. `Fail`, `DeadLetter` and `Reschedule` each point at it.

Sanitizing at the endpoint was the alternative, but it cannot tell safe text from unsafe once it is in the column — the distinction only exists at the call site. Since no runner exists yet, no writer has been built against the wrong assumption, which makes now the cheapest moment to state it. Adding it to the J3 checklist so the runner is written against a stated rule rather than inferring one.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
igdragoev: Fixed — `summary.Responses[400]` and `.ProducesProblem(400)` both added. The generated document now declares 200, 400, 401 and 404.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
igdragoev: Agreed, and fixed — the handler takes an `ILogger` and logs a warning with the job id when the payload will not parse. Reporting no result is still the behaviour; it is only the silence that changes.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
igdragoev: All four, and two of them were real defects in the documentation rather than tidying.

`FirstOrDefaultAsync` on the primary-key lookup, and `AsNoTracking` dropped from a projection that was never tracked.

The comment about a failed job carrying an earlier result described a state the machine cannot reach — `Complete` is the only writer of `ResultJson` and it is terminal. Reworded to say the gate is defence against a future writer, so the next reader is not sent looking for a state that does not exist. The gate stays.

The `ErrorMessage` documentation was the one that would have cost someone real time: `Reschedule` sets it with `Status = Retrying`, so a client following "set only once it has failed" would have stopped polling a job that was merely between attempts. It now says the message describes the last attempt rather than the job, and points at `Status` as the authoritative signal — same wording on the entity property.

`ExecuteAsync_UnfinishedJob_ReportsNoResult` is deleted rather than strengthened. You are right that it asserted `FromDto` copies a null; the gate it claimed to cover is in the handler, and `Handle_UnfinishedJob_ReportsNoResult` already tests it where it lives.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
```diff
@@ -0,0 +1,78 @@
...
+/// A job carries no data of its own beyond its progress — the work it did is reached through
+/// whatever endpoint owns that work, which applies its own permissions there.
+/// </remarks>
+public sealed class GetJob(IMediator mediator, ITenantContext tenantContext)
+    : Endpoint<GetJobRequest, Results<Ok<JobResponse>, ProblemHttpResult>>
+{
+    /// <inheritdoc />
+    public override void Configure()
+    {
+        Get("jobs/{jobId}");
```
PeterMarinov, line 30: **No `Permissions(...)` after the verb.**

Default policy is authenticated + `sub`. This is not self-scoped (`Auth/Me` / logout / change-password). Any tenant member can poll any job, including system-enqueued rows.

Interim (no jobs action in the catalog yet):

```csharp
Get("jobs/{jobId}");
Permissions(Actions.Tenant.ViewSettings);
```

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
```diff
@@ -0,0 +1,78 @@
...
+                "Retrieves the current state of a background job. Progress and status message are " +
+                "advisory and may lag the work itself; status is authoritative.";
+            summary.ExampleRequest = new GetJobRequest { JobId = 987654321 };
+            summary.Responses[200] = "Job state retrieved.";
+            summary.Responses[400] = "The job ID is not valid.";
+            summary.Responses[404] = "Job not found.";
+        });
+        Description(builder => builder
+            .Produces<JobResponse>(200, "application/json")
+            .ProducesProblem(400)
```
PeterMarinov, line 46: **Application `401` is a first-class status here and is not declared.**

`GetJobHandler` returns `Result.Unauthorized("Tenant context is required.")` → `ToProblem` 401. `Configure` only lists 200 / 400 / 404.

`oss/AGENTS.md` → Error HTTP contract / OpenAPI: `Description` must list **exactly** the statuses `Summary.Responses` declares; every `ProblemHttpResult` path needs `ProducesProblem`. Auth-scheme 401 ≠ application 401 with a problem body.

`.cursor/rules/endatix-api-rules.mdc` → OpenAPI rules: always document errors with `ProducesProblem`; never invent extras — this 401 is real, not invented.

Reference for declaring 401: `Endpoints/Auth/AssumeTenant.cs` (`summary.Responses[401]` + `.ProducesProblem(StatusCodes.Status401Unauthorized)`).

Also add `GetJobEndpointTests`: `Result.Unauthorized` → `ProblemHttpResult` 401 (`GetActiveTests` / `oss/AGENTS.md` FastEndpoints error contract).

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.JobResponse.cs
```diff
@@ -0,0 +1,55 @@
+using Endatix.Core.Abstractions.BackgroundJobs;
+using Endatix.Modules.Jobs.Features.GetJob;
+
+namespace Endatix.Modules.Jobs.Endpoints.Jobs;
+
+/// <summary>
+/// The state of a background job.
+/// </summary>
+public sealed class JobResponse
```
PeterMarinov: **Split response file — collocate with the endpoint.**

`GetJob.cs` already holds the endpoint, request, and validator. `JobResponse` in a sibling file is extra for this size.

**Source of truth (same file):** `.cursor/rules/endatix-api-rules.mdc` → Project Structure → Endatix.Api:

> `[Operation].cs`: Main endpoint implementation. **In the same file** add the Request, Response and Validators classes as required `[Operation][Request/Response/Validator]`

Also Validation Rules: “Prefer using the same files as the endpoints using `[Request]Validator` as class name.”

**In-file example:** `Endpoints/Forms/List.cs` — `List` + `FormsListValidator` + `FormsListRequest` in one file.

**Configure / Permissions / OpenAPI example** (requested): `Endpoints/Forms/Delete.cs` — verb, `Permissions(...)`, `Summary` + `Produces` / `ProducesProblem`. (`Delete` uses nested `Delete.DeleteForm*.cs` for request/validator; do **not** copy that split for this endpoint.)

Move `JobResponse` (+ `FromDto`) into `GetJob.cs`; delete `GetJob.JobResponse.cs`. Trim XML that restates `JobDto` remarks.

src/Endatix.Modules.Jobs/Features/GetJob/GetJobHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+    : IQueryHandler<GetJobQuery, Result<JobDto>>
+{
+    public async Task<Result<JobDto>> Handle(GetJobQuery request, CancellationToken cancellationToken)
+    {
+        // The ambient tenant filter cannot carry this on its own. It reads a tenant id of zero as
+        // "no tenant, show everything" so that background services can sweep across tenants — and a
+        // request whose principal has no usable `tid` claim also leaves the tenant context at zero,
+        // because TenantMiddleware returns early rather than failing. Scoping such a request by the
+        // filter alone would serve it every tenant's jobs.
+        if (request.TenantId <= 0)
```
PeterMarinov, line 49: **Tenant-0 early-return is the security fix; CI does not cover the host path.**

Ambient filter is `GetTenantId() == 0 || e.TenantId == GetTenantId()`. Tenant `0` is **bypass, not isolation**.

Refs:
- `oss/ARCHITECTURE.md` → Multi-tenancy → Data isolation: “Ambient tenant is `ITenantContext.TenantId` … **Tenant `0` is bypass, not isolation**.”
- `.cursor/rules/endatix-api-rules.mdc` → Query filters: same rule; HTTP work must pass `tenantId` explicitly; never use tenant `0` as the “other” tenant in a test.

This `if (request.TenantId <= 0)` is correct. `Handle_NoTenantOnTheRequest` never hits EF. InMemory + `TestJobsDbContext` is not `TenantMiddleware` + PostgreSQL.

`.cursor/rules/endatix-api-rules.mdc` → Unit vs integration: HTTP + auth + middleware + EF → integration (`WebApplicationFactory` / Testcontainers). Do **not** use in-memory for endpoint isolation coverage.

Prefer one FeatureFlows/CriticalPaths HTTP case with Jobs flag on: valid `tid` → 200; same user without `tid` → 401; other-tenant id → 404. If the suite cannot enable the flag yet, say so in the PR — do not claim the filter is exercised.

The class remark that the query filter is “the only thing standing between tenants” is stale; the `Where(TenantId == request.TenantId)` is.

src/Endatix.Modules.Jobs/Domain/BackgroundJob.cs
```diff
@@ -83,7 +83,18 @@ public BackgroundJob(
...
+    /// Why the most recent attempt failed. Set on <see cref="JobStatus.Failed"/> and
+    /// <see cref="JobStatus.DeadLettered"/>, and also on <see cref="JobStatus.Retrying"/>, where it
+    /// describes the attempt just ended rather than the job.
+    /// </summary>
+    /// <remarks>
+    /// <b>Must be author-written text that is safe to show an end user.</b> The job status endpoint
+    /// returns this column verbatim, so an exception message assigned here is published to the
+    /// caller — and exception text from EF Core, Npgsql and the BCL carries connection strings, SQL
+    /// and file paths. Log the exception and correlate by <see cref="TraceId"/>; put a sentence the
+    /// requester can act on here.
```
PeterMarinov: **`ErrorMessage` safety is XML, not a guard.**

This endpoint returns the column verbatim. `ResultFactoryMustNotInterpolateExceptionMessageTests` does not match `job.Fail(ex.Message, …)`.

`.cursor/rules/endatix-api-rules.mdc` / `oss/AGENTS.md` → Exception text never reaches the caller: `ex.Message` must not land in client-visible `detail`. Same rule applies once this column is on the wire.

Keep **one** sentence on the property. Drop the duplicated remarks on `Fail` / `Reschedule` / `DeadLetter`. Enforce on the first writer (J3), not with more XML.

PeterMarinov (changes requested):
Great, this is almost prod ready. Comments are related to align with recent effort to move to single file endpoints and to add proper permission for the endpoint. Let's not merge before addressing these changes. Summary:

**Must**
- `Permissions(Actions.Tenant.ViewSettings)` - to protect the endpoint from public anonymous access. We don't have background jobs-specific permission yet; this is the interim.
- One file: endpoint + request + response + validator (`GetJob.JobResponse.cs` folds into `GetJob.cs`) - this is stylistic, but helps with leaners code base and better cohesion
- Declare and test the application `401` (`tid` missing).

**Should**
- HTTP integration (or an explicit defer) for tid-missing / cross-tenant. InMemory does not stand in for `TenantMiddleware` + PG filters.
- Short ARCHITECTURE/AGENTS note: HTTP reads pass `TenantId` and refuse `<= 0`; module endpoints still call `Permissions(...)`.

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
igdragoev: Fixed

src/Endatix.Modules.Jobs/Endpoints/Jobs/GetJob.cs
igdragoev: I added Actions.Jobs.View permission. It is up to is to decide which roles will get this permission.

PeterMarinov (dismissed):
Approved as it is but will be best if we seed the rest of the anticipated Jobs granular permissions so we don't have to add more permissions for the upcoimign endpints

PR #1026, opened 2026-09-09
fix(e1015): Ignore tenant filters when seeding default export formats

PeterMarinov:
## Summary
- `SeedDefaultsAsync` now ignores EF tenant query filters and keys off the explicit `tenantId` (plus `!IsDeleted`).
- Needed because outbox `tenant.created` is app-level (`Tenant` is not `ITenantOwned`, row `TenantId = 0`).
- Slim absorb of [#1015](https://github.com/endatix/endatix/issues/1015) / alternative to [#1016](https://github.com/endatix/endatix/pull/1016): no value-generator rewrite, no race/soft-delete machinery. Id=0 `AddRange` is already avoided on `feature/e1012-xlsx-export` (one save per row).
- Export formats seeder hardening - added use of IUniqueConstraintViolationChecker + fix for stale default mapping + use of Named EF Core query filters tests

## Test plan
- [x] `ExportFormatSeedIntegrationTests.SeedDefaultsAsync_WhenAmbientTenantDiffers_SeedsAndDoesNotDuplicate`
- [x] Existing seed tests still pass

Closes #1015 (with parent stack). Prefer this over merging #1016.

## Summary by CodeRabbit

* **Improvements**
  * Strengthened tenant data isolation across application and reporting queries.
  * Improved default export format setup to prevent duplicate records and recover gracefully when formats are deleted or already exist.
  * Default mappings now reconnect to available export formats when needed.

* **Documentation**
  * Added guidance on tenant-aware data filtering and background processing.
  * Updated integration testing instructions and formatting.


src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
```diff
@@ -273,16 +263,65 @@ private async Task EnsureDefaultMappingAsync(long tenantId, CancellationToken ca
...
+    private async Task SaveOrConcedeAsync(EntityEntry entry, CancellationToken cancellationToken)
+    {
+        try
+        {
+            await unitOfWork.SaveChangesAsync(cancellationToken);
+        }
+        catch (DbUpdateException exception)
+            when (uniqueViolationChecker.AnalyzeUniqueConstraint(exception).IsUniqueConstraintViolation)
+        {
+            entry.State = EntityState.Detached;
```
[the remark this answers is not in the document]
PeterMarinov: This is addressed, please check

igdragoev (approved):

src/Endatix.Modules.Reporting/Persistence/Migrations/PostgreSql/20260904194346_SeedDefaultExportFormats.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Backfills Native CSV, JSON, Excel, and Codebook (plus CSV default mapping) for every tenant.
+    /// SQL comes from <see cref="DefaultExportFormats"/>. Ids follow the <c>InitialReporting</c>
+    /// <c>hashtextextended</c> convention.
+    /// </summary>
+    public partial class SeedDefaultExportFormats : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(DefaultExportFormatsPostgresBackfill.UpSql);
```
igdragoev: **`Up` executes SQL generated at runtime from a mutable catalog.**

`DefaultExportFormatsPostgresBackfill.UpSql` is built from `DefaultExportFormats.All`, so this migration's SQL changes whenever the catalog does — but `__EFMigrationsHistory` has already recorded `20260904194346`, so applied databases never re-run it.

Append a fifth default later and the two halves diverge permanently: existing tenants stay at four formats with **no migration able to fix it**, while a freshly-migrated database gets five from the same migration id. `dotnet ef migrations script` also emits different SQL for the same id between builds, so generated deploy scripts stop matching what was actually applied.

Migration bodies need to be immutable literals. Inline the SQL as it stands today and let the catalog drive only the runtime seeder — otherwise the catalog and the migration history are two sources of truth for the same rows.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
```diff
@@ -199,29 +175,27 @@ await dbContext.SurveyTypeExportMappings
...
-        var missing = DefaultFormats.Where(definition => !existing.Any(format =>
+        var missing = DefaultExportFormats.All.Where(definition => !existing.Any(format =>
             format.ExportTarget == definition.Target &&
             format.DeliveryFormat == definition.Delivery));
 
-        // One save per row: BaseEntity.Id is DatabaseGeneratedOption.None, so EF assigns no
-        // temporary keys and ReportingDbContext stamps the id on save. Adding several unsaved
-        // rows at once would put two Id = 0 entities in the change tracker and throw.
+        // One save per row: BaseEntity.Id is DatabaseGeneratedOption.None; two unsaved rows collide on Id = 0.
         foreach (var definition in missing)
```
igdragoev, line 189: **One name collision aborts the whole seed, and the tenant can never recover.**

If a tenant already has a format named `CSV` with a non-Native profile (say Shoji), it is not in `existing`, so `CSV` stays in `missing`. The insert hits 23505 on `IX_ExportFormats_TenantId_Name`, `IsSameDefaultAsync` returns false, and this rethrows — so JSON, Excel and Codebook are never attempted and `EnsureDefaultMappingAsync` never runs.

The picker lists `reporting.ExportFormats` rows rather than registered capabilities, so that tenant ends up with **zero usable export formats and no default mapping**. Because every at-least-once redelivery of `tenant.created` fails identically, the state is permanent rather than transient.

`SeedDefaultsAsync_WhenNameTakenByNonNativeCsv_ThrowsAndDoesNotCreateDefaultMapping` pins this as intended, so it is a deliberate choice worth revisiting: skipping the colliding definition and continuing would leave the tenant usable, and the collision is already visible to the operator.

igdragoev (commented):
Three findings from a review of this PR. The core fix — opting out of the *named* `Tenant` filter instead of relying on ambient tenant 0 — is right and well documented. These are all in code the PR added beyond that fix, and I'd treat them as merge blockers: the first changes what an already-applied migration means, and the other two are reachable through ordinary `tenant.created` outbox redelivery.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
```diff
@@ -273,16 +229,97 @@ private async Task EnsureDefaultMappingAsync(long tenantId, CancellationToken ca
...
-            surveyTypeId: null,
-            isDefault: true);
+        var defaultMapping = await MappingsForTenant(tenantId)
+            .FirstOrDefaultAsync(
+                mapping => mapping.IsDefault && mapping.SurveyTypeId == null,
+                cancellationToken);
+
+        if (defaultMapping is null)
+        {
+            var entry = dbContext.SurveyTypeExportMappings.Add(
```
igdragoev, line 237: **This inserts a second mapping for the same scope once the tenant default has been deliberately cleared.**

Seeding creates `(tenant, SurveyTypeId=null, csvId, IsDefault=true)`. An admin then calls `ExportMappingRepository.UpsertAsync` with `isDefault: false`, which `ClearDefault()`s that row. On the next `tenant.created` redelivery the `IsDefault && SurveyTypeId == null` lookup finds nothing, so a second `(tenant, null, csvId)` row is added with `IsDefault = true`.

Neither unique index rejects it — `IX_SurveyTypeExportMappings_TenantId` and its typed sibling are both filtered on `"IsDefault" = true`, and the cleared row no longer matches. `SurveyTypeExportMapping`'s own doc comment says *"Duplicate format links per scope are enforced in application code"*, but nothing on this path enforces it. `ListAsync` then returns two identical-scope entries and `UpsertAsync`'s `FirstOrDefaultAsync` picks one arbitrarily, so the default toggle becomes unreliable.

Looking the mapping up by `(TenantId, SurveyTypeId)` regardless of `IsDefault`, then re-marking it, avoids the duplicate and also restores the admin's intent explicitly rather than by accident.

src/Endatix.Modules.Reporting/Persistence/Migrations/PostgreSql/20260904194346_SeedDefaultExportFormats.cs
PeterMarinov: Agreed — migration bodies have to be immutable. Up/Down now use frozen SQL in SeedDefaultExportFormatsSql, not DefaultExportFormats.All. The catalog only drives SeedDefaultsAsync / tenant.created. A later default gets its own data migration so __EFMigrationsHistory stays one source of truth.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
PeterMarinov: Agreed — failing the whole seed was the wrong trade-off. A name clash with a non-Native row is skipped now; JSON, Excel, and Codebook still insert. We still don’t create Native CSV (or a default mapping) under that taken name. tenant.created redelivery can succeed. Test updated: remaining defaults present, no Native CSV, no default mapping.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
PeterMarinov: Agreed — filtered unique indexes don’t stop a second tenant-scope row after ClearDefault(). Seed now loads that scope without IsDefault, and won’t insert again. We don’t flip IsDefault back on outbox redelivery (that would undo the admin). Dead-format PointTo only runs if the mapping is still the tenant default. Frozen migration SQL uses the same SurveyTypeId IS NULL existence check.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
```diff
@@ -273,16 +229,91 @@ private async Task EnsureDefaultMappingAsync(long tenantId, CancellationToken ca
             return;
         }
 
-        SurveyTypeExportMapping defaultMapping = new(
-            tenantId,
-            csvFormatId.Value,
-            surveyTypeId: null,
-            isDefault: true);
+        var defaultMapping = await MappingsForTenant(tenantId)
```
igdragoev, line 232: **A soft-deleted default mapping makes this silently give up, permanently.**

The two indexes are asymmetric. `ExportFormatConfigurationPostgreSql` filters on `"IsDeleted" = false`, so a soft-deleted format frees its index slot and can be re-seeded — which is why `SeedDefaultsAsync_WhenFormatSoftDeleted_ReseedsIt` passes. `SurveyTypeExportMappingConfigurationPostgreSql` does **not**: both its unique indexes are filtered on `"IsDefault" = true` alone, so a soft-deleted row still occupies the index.

With a soft-deleted `(tenant, SurveyTypeId = null, IsDefault = true)` mapping present:

1. `MappingsForTenant` keeps the SoftDelete filter by design, so the row is invisible → `defaultMapping is null`.
2. The insert at :237 raises 23505 against `IX_SurveyTypeExportMappings_TenantId`.
3. `SaveOrConcedeAsync` catches it and asks `IsSameDefaultAsync`, whose mapping branch (:295) queries through **the same** SoftDelete filter, finds nothing, and returns false.
4. `if (!false)` → `return`. Swallowed.

So the mapping is never created, `GetTenantDefaultAsync` returns null for that tenant forever, and every `tenant.created` redelivery repeats the same no-op. There is no exception and no log line, which makes this harder to diagnose than the wedge it replaced — the tenant simply has no default export format.

Adding `AND "IsDeleted" = false` to both mapping index filters lines them up with `ExportFormats` and makes the existing re-seed behaviour apply here too. Failing that, the collision probe in `IsSameDefaultAsync` needs a bare `IgnoreQueryFilters()` so it can see the row that actually caused the violation, and undelete or repoint it.

igdragoev (commented):
One follow-up in the same area, found while re-reading the updated `EnsureDefaultMappingAsync`. The three earlier points look properly addressed — and the mapping fix is better than what I suggested: declining to re-flip `IsDefault` preserves a deliberate `ClearDefault()`, which my "re-mark it" would have undone.

src/Endatix.Modules.Reporting/Data/ExportFormatRepository.cs
PeterMarinov: Valid finding. For some reason the unique indexes did not include the "IsDeleted" - false. Indexes updated + the migration, so we can merge it. Tests added as well

igdragoev (approved):

PR #1030, opened 2026-09-09
chore(e1029): share a reusable tenant context stub across integration tests #1029

### LINKED ISSUE #1029: Share a reusable tenant context stub across integration tests
### Goal

Give integration tests one reusable `ITenantContext` stub in `Endatix.IntegrationTests.Shared` instead of hand-rolling it per test class, so tenant-scoped tests read the same way everywhere.

### Implementation Notes

Two identical private classes already exist inside `Endatix.IntegrationTests`:

- `ReportingTestSchema.BypassTenantContext`
- `IntegrationCoreMigrationTestHelper.BypassTenantContext`

plus ~15 `Substitute.For<ITenantContext>()` + `.TenantId.Returns(id)` pairs across the reporting and form-schema integration tests.

Add a single `TestTenantContext(long tenantId) : ITenantContext` to the shared project and use it in those places.

Constraints that shape the change:

- `Endatix.IntegrationTests.Shared` ships as the **`Endatix.IntegrationTesting`** NuGet package (`IsPackable=true`), so anything added there becomes supported public API. Keep the surface to this one type.
- The stub must be a plain hand-written class, **not** a `Substitute.For<>` factory: the csproj marks NSubstitute `PrivateAssets="all"` precisely so it does not become a consumer dependency.
- Keep the shared project **module-agnostic** — per `tests/Endatix.IntegrationTests/AGENTS.md`, module helpers such as `ReportingTestSchema` stay in `Endatix.IntegrationTests`. Reporting-specific context/repository factories are explicitly out of scope.

Tenant `0` means *bypass* rather than isolation, so the type should make that legible at call sites (see `ARCHITECTURE.md` → Multi-tenancy → "Data isolation (EF query filters)").

Stacked on #1026; branch from `chore/e1015-absorb-export-format-id-collision` until that merges.

PeterMarinov:
## Summary

Shared **`TestTenantContext`** for OSS integration tests that construct a `DbContext` directly, instead of a per-file NSubstitute `ITenantContext`.

Closes #1029.

## What this ships

- `Endatix.IntegrationTests.IntegrationTenantContext` (this test project, **not** the `Endatix.IntegrationTesting` package).
- `new(tenantId)` for a scoped tenant — **rejects 0 / negative**.
- `TestTenantContext.Bypass` (`TenantId = 0`) to turn the EF tenant filter off on a hand-built DbContext.
- Call sites in Reporting and related DB-only tests switched to the shared type.

## APIs / events

None. Test infrastructure only. Not a NuGet surface.

## Key decisions

- Tenant `0` is **filter bypass**, not “another tenant”. Isolation tests use two non-zero ids (`ReportingQueryFilterTests`).
- `Bypass` is for DbContext construction only. Core handlers `Guard` / Invalid on tenant 0 — do not inject Bypass into those.
- Nested `TestTenantContext` stubs in Api.Tests / Infrastructure.Tests stay local; those projects do not reference this helper.
- Substitute remains only where `TenantId` is reassigned mid-test (`OutboxCaptureTests`).

## Behavior / compatibility

- No production behavior change.
- Tests that previously used `Substitute.For<ITenantContext>()` with a fixed non-zero id should behave the same.
- Accidental `new TestTenantContext(0)` now throws instead of silently seeing every tenant.


tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
```diff
@@ -0,0 +1,14 @@
+using Endatix.Core.Abstractions;
+
+namespace Endatix.IntegrationTests.Shared;
+
+/// <summary>
+/// Fixed-tenant <see cref="ITenantContext"/> for tests that construct a DbContext directly.
+/// Tenant 0 (<see cref="Bypass"/>) turns the tenant filter off — never use it as the "other" tenant.
+/// </summary>
+public sealed class TestTenantContext(long tenantId) : ITenantContext
```
igdragoev: **Tenant 0 is load-bearing here, but nothing enforces it.**

The doc comment two lines up states the invariant — tenant 0 turns the filter off, never use it as the "other" tenant — and then the primary constructor accepts 0 like any other value.

The failure is quiet. A test does `long tenantId = someRecord.TenantId;` where that field is unset, or a helper defaults a parameter, and `new TestTenantContext(0)` produces a context that sees *every* tenant's rows rather than an empty scope: `ApplyEndatixQueryFilters` emits `currentTenantId == 0 || e.TenantId == currentTenantId` (`src/Endatix.Infrastructure/Data/DbContextModelBuilderExtensions.cs:49-58`). An isolation assertion like `(await ctx.ExportFormats.ToListAsync()).Should().ContainSingle()` then passes or fails for reasons that have nothing to do with isolation, and a genuine tenant-leak regression goes unnoticed.

This is exactly the kind of rule worth making unrepresentable rather than documenting: `Guard.Against.NegativeOrZero(tenantId)` in the primary constructor, with `Bypass` constructed through a private path. The type is small enough that the change is a few lines, and it turns a silent wrong-answer into an immediate, obvious failure.

tests/Endatix.IntegrationTests.Shared/README.md
```diff
@@ -20,6 +20,7 @@ This is a test-only package — reference it from a test project, not from produ
 - **`IIntegrationTestHostFixture`** — a host fixture over `WebApplicationFactory` with authentication wired up.
 - **`TestPersona`** / **`IntegrationAuthClients`** — authenticated callers for permission and tenant-isolation tests.
 - **`DatabaseCheckpoint`** — Respawn-based reset so each test starts from a known state.
+- **`TestTenantContext`** — a fixed-tenant `ITenantContext` for tests that build a DbContext directly, with `TestTenantContext.Bypass` for app-level access that sees every tenant.
```
igdragoev: **"app-level access that sees every tenant" points a consumer at a use that always throws.**

It is accurate about the query filter, but it reads as a general-purpose escape hatch, and this is the `PackageReadmeFile` — the first and often only documentation a consumer of `Endatix.IntegrationTesting` sees.

A reader who takes "app-level access" at face value and wires `TestTenantContext.Bypass` into a test that exercises a Core handler gets an opaque guard failure: `CreateFolderHandler`, `CreateFormHandler`, `CreateCustomQuestionHandler`, `GetTenantSettingsHandler` and `ListFoldersHandler` all open with `Guard.Against.NegativeOrZero(tenantContext.TenantId)`, and `CreateDataListHandler` returns Invalid on `<= 0`. Tenant 0 is not a privileged caller anywhere above the DbContext — it is a value the application layer actively rejects.

The XML doc and the AGENTS.md entry both get this right by scoping the type to "tests that construct a DbContext directly". Worth saying the same here, e.g. "...with `TestTenantContext.Bypass` to disable the tenant query filter when a test needs to read across tenants directly from a DbContext".

tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static TestTenantContext Bypass { get; } = new(0);
```
igdragoev: **This commits tenant-0 bypass to a shipped package's public surface for a single in-repo caller.**

The csproj sets `IsPackable=true` / `PackageId=Endatix.IntegrationTesting`, and this PR's own AGENTS.md addition warns that anything added there becomes public API. All 13 call sites are inside `tests/Endatix.IntegrationTests`, and `Endatix.SaaS.IntegrationTests` — the only other consumer — has no `ITenantContext` references at all.

What gets locked in is not just a type but a semantic: that `Bypass` means tenant 0, and that tenant 0 disables the filter. That second half is an implementation detail of `ApplyEndatixQueryFilters`. If the filter ever grows a real "all tenants" sentinel, or the bypass moves behind an explicit `IgnoreQueryFilters` convention, this member has to keep meaning what it means today for external consumers.

Given there is no consumer outside this repo, `tests/Endatix.IntegrationTests/Infrastructure/` alongside `IntegrationAppDbContextFactory` seems the better home until a second one appears — or `internal` plus `InternalsVisibleTo` if it should stay in Shared. Either keeps the option open; publishing does not.

tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
```diff
@@ -0,0 +1,14 @@
+using Endatix.Core.Abstractions;
+
+namespace Endatix.IntegrationTests.Shared;
```
igdragoev: **`TestTenantContext` is the conventional name for the private stub in sibling test projects, so exporting it at namespace scope sets up a collision.**

Three projects already declare their own: `tests/Endatix.Api.Tests/Common/FeatureFlags/FeatureGateTargetingTests.cs:111` (character-identical to this type), `FeatureGateTests.cs:75`, and `tests/Endatix.Infrastructure.Tests/FeatureFlags/FeatureFlagsTargetingContextTests.cs:56`. They are nested classes today, so nothing breaks right now.

The break arrives indirectly: a project with a namespace-scope `TestTenantContext` that later adds `using Endatix.IntegrationTests.Shared;` to pick up `DbIntegrationFixture` gets `error CS0104: 'TestTenantContext' is an ambiguous reference` on every existing use — caused entirely by importing an unrelated fixture.

A package-scoped name — `FixedTenantContext` or `IntegrationTenantContext` — sidesteps it without giving up anything.

Related, and worth deciding alongside this: the new AGENTS.md rule "do not hand-roll an `ITenantContext` stub" cannot be followed by those three files, since neither `Endatix.Api.Tests` nor `Endatix.Infrastructure.Tests` references `Endatix.IntegrationTests.Shared`. As it stands the PR yields a fourth copy of the same type rather than one. Either scope the rule to `Endatix.IntegrationTests` explicitly, or put the type somewhere all four projects can reach.

igdragoev (commented):
### Review — shared `TestTenantContext`

No correctness problems: the substitution is faithful, and the tenant-0 bypass semantics hold for every affected context (`AppDbContext`, `ReportingDbContext` and SaaS `AgentsDbContext` all route through `ApplyEndatixQueryFilters`). Four comments below, all about the new type rather than the call sites it replaces.

Two of them — the naming one and the package-surface one — are really a single decision: if `TestTenantContext` moves into `tests/Endatix.IntegrationTests/Infrastructure/` next to `IntegrationAppDbContextFactory` until a second consumer exists, both stop applying.

tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
PeterMarinov: Fixed. The public constructor now guards:
 
 ```cs
 Guard.Against.NegativeOrZero(tenantId, nameof(tenantId),
      $"Tenant id must be positive. Use {nameof(IntegrationTenantContext)}.{nameof(Bypass)} to disable the tenant filter.");
 ```     
- Guard.Against rather than a hand-written throw, to match JwtTokenService / the command constructors. 

Tenant 0 is now reachable only through the private ctor behind Bypass, so new IntegrationTenantContext(someUnsetTenantId) fails loudly instead of silently seeing every tenant. Pinned by IntegrationTenantContextTests: 0 and -1 throw, Bypass.TenantId is 0.

tests/Endatix.IntegrationTests.Shared/README.md
PeterMarinov: Reverted - the line is out of the Shared PackageReadmeFile entirely. The type no longer lives in Shared, so there is nothing here for a package consumer to read.
 
 Bypass semantics are now stated only where they apply, and scoped: the XML doc on Bypass ("DbContext construction only: Core handlers guard against tenant 0") and the tests/Endatix.IntegrationTests/AGENTS.md bullet ("turns the query filter off for DbContext reads only — Core handlers reject it").

tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
PeterMarinov: Agreed. Moved to `tests/Endatix.IntegrationTests/Infrastructure/IntegrationTenantContext.cs`, alongside `IntegrationAppDbContextFactory. Shared` is untouched by this PR - nothing new lands in `Endatix.IntegrationTesting`, so Bypass == 0 stays an in-repo convention we can change with the filter rather than a public contract.

Recorded in AGENTS.md so it does not regress: module helpers live in IntegrationTests, not Shared, because Shared ships as a package and its NSubstitute is PrivateAssets="all". Promote it if a second consumer appears.

tests/Endatix.IntegrationTests.Shared/TestTenantContext.cs
PeterMarinov: Renamed to `IntegrationTenantContext`. No ambiguity now with the nested stubs in FeatureGateTests, FeatureGateTargetingTests and FeatureFlagsTargetingContextTests, whichever fixtures those projects import later.

On the rule: it moved with the type into tests/Endatix.IntegrationTests/AGENTS.md and is scoped to this project - "Substitute an ITenantContext only where TenantId changes mid-test (OutboxCaptureTests)". It no longer makes a claim over Api.Tests / Infrastructure.Tests, which cannot reach the helper.

Worth noting for a follow-up: Endatix.Modules.Jobs.Tests/Shared/TestDoubles.cs already has FixedTenantContext for the same concept, so the repo is at three names across four projects. Consolidating is a separate cleanup - happy to file it.

igdragoev (approved):

PR #1044, opened 2026-09-14
fix(#1049): stop exception text reaching a background job's ErrorMessage

### LINKED ISSUE #1049: Guard background job error messages against exception text
Related: #1002. It applies the same rule to `Result` factory details, but its guard only covered those factories.

### Goal

A background job's `ErrorMessage` must be author-written text that is safe to show an end user. `GET jobs/{jobId}` returns it verbatim, so any exception text stored there reaches the caller, and EF Core, Npgsql and BCL exception messages carry connection strings, SQL and file paths.

### Implementation Notes

- Extend `ResultFactoryMustNotInterpolateExceptionMessageTests` to flag exception text passed to `Fail`, `Reschedule`, `DeadLetter` or `RecordFailedAttemptAsync`, and in a `SetProperty` setter that targets `ErrorMessage`.
- Flag only the exception-message argument, not the method name, so unrelated `Fail` methods (`AuthorizationHandlerContext.Fail`, `ValidateOptionsResult.Fail`) do not trigger it.
- Nothing writes the column yet. This guard lands before the first writer does.

Out of scope: the job runner itself. When a handler throws, the runner must resolve the message with `SafeError.LogAndResolve` and log the exception against the job id and `TraceId`. That requirement is tracked in the background jobs plan's runner checklist (#862).

igdragoev:
Closes #1049

## Description of changes

`GET jobs/{jobId}` returns a job's `ErrorMessage` verbatim, which makes whatever records a job's failure as client-visible as a `Result` factory. The exception-text guard only knew about `Result` factories, so the obvious runner code — `job.Reschedule(next, ex.Message)` — would pass it and publish whatever the exception carried. EF Core, Npgsql and BCL messages carry connection strings, SQL and file paths.

`ResultFactoryMustNotInterpolateExceptionMessageTests` now also flags exception text passed to:

- `Fail`, `Reschedule`, `DeadLetter` — the job state machine's failure writers
- `RecordFailedAttemptAsync` — the repository call that records a failed attempt
- a `SetProperty` setter aimed at `ErrorMessage` — set-based updates write the column without going through the entity

Only the exception-message argument is flagged, never the call, so unrelated methods named `Fail` — `AuthorizationHandlerContext.Fail(...)`, `ValidateOptionsResult.Fail(...)`, both already in `src` — stay clean.

Nothing writes the column yet. The runner will be its first writer, and it must route a thrown handler's message through `SafeError.LogAndResolve`; this puts the guard in place before that code exists.


PeterMarinov (approved):
Approved.

I can see same trust boundary as `Result.*`. When the runner lands, resolve via `SafeError.LogAndResolve` - so reusing the app error filtering path, which is great

Issue #1051, opened 2026-09-15
Run background jobs with retry, crash recovery and cancellation


igdragoev:
### Goal

Make the Jobs module execute the jobs it enqueues. A bounded runner claims each job, runs its handler, keeps a heartbeat, and records retry, failure, dead-letter or cancellation on the job row. A sweeper on every instance hands back jobs whose heartbeat stopped, rediscovers eligible jobs, and warns when eligible jobs wait unclaimed too long, so the queue needs no second system. The module stays behind `Endatix:FeatureFlags:JobsModule`, off by default.

### Implementation Notes

- Delivered as a stack of PRs linked to this issue. The module flag stays off throughout.
- Settings live under `Endatix:BackgroundJobs` and are validated at startup. `MaxAttempts`, `MaxRuntimeMinutes`, `BackoffBaseSeconds` and `BackoffCapSeconds` can be overridden per job type. Retries use capped exponential backoff.
- Job state writes are set-based and fenced on the claimed attempt, through an internal `IBackgroundJobStateRepository`.
- `AttemptCount` increments at claim (#1017). Host shutdown leaves a job `Processing`, and the sweeper re-queues it once its heartbeat is stale.
- A thrown handler's message reaches `ErrorMessage` only through `SafeError.LogAndResolve` (#1049).
- The job row carries state only. The unused `ResultJson` column and the `resultJson` parameter of `BackgroundJob.Complete` are removed; a handler that produces output saves or publishes it itself. This breaks the published `Endatix.Modules.Jobs` 0.7.6 surface, which has no consumers.
- Dispatch strategy and metrics are public extension points (#1064). Their defaults are registered with `TryAdd*` together with the runner and sweeper, so a replacement from another assembly still wins. The metrics default records on the `Endatix.Jobs` meter, which the host's OpenTelemetry pipeline exports.

Out of scope: webhook delivery through jobs and job retention, submission export and download, recurring jobs, and requeueing dead-lettered jobs.

PR #1052, opened 2026-09-15
feat(#1051): bind background job options and the retry backoff policy

igdragoev:
Part of #1051

## Description of changes

The job runner and sweeper need their settings before they exist. This adds the settings, their validation and the retry backoff. Nothing registers them yet, so a running host is unchanged.

- **`BackgroundJobsOptions`** binds `Endatix:BackgroundJobs`. `MaxAttempts`, `MaxRuntimeMinutes`, `BackoffBaseSeconds` and `BackoffCapSeconds` can be overridden per job type under `JobTypes:{JobType}`. Each key falls back to the global value on its own. Override keys match the job type regardless of case, because configuration keys are case-insensitive everywhere else; a dictionary assigned in code is copied into a case-insensitive one.
- **`BackgroundJobsOptionsValidator`** rejects values jobs cannot run with, naming the full configuration key:
  - any value below 1;
  - a stuck-job threshold shorter than three heartbeats, which would presume a slow but healthy job lost;
  - a backoff cap below its base.

  Unknown job types and sibling sections pass, because an override for a handler that is not deployed yet is not a mistake.
- **`BackgroundJobRetryPolicy.NextAttemptAt`** waits the base delay after the first failed attempt, doubles it for each later attempt, and stops at the cap. The attempt count increments at claim, so a first failure is already attempt 1. An attempt below 1 is treated as the first.

| Key | Default |
|---|---|
| `RunInProcess` | `true` |
| `MaxConcurrency` | 4 |
| `SweepIntervalSeconds` | 10 |
| `SweepBatchSize` | 200 |
| `StuckJobThresholdMinutes` | 10 |
| `HeartbeatIntervalSeconds` | 30 |
| `MaxRuntimeMinutes` | 60 |
| `MaxAttempts` | 3 |
| `BackoffBaseSeconds` | 30 |
| `BackoffCapSeconds` | 900 |
| `BacklogWarningMinutes` | 15 (global only) |

## Summary by CodeRabbit

- **New Features**
  - Added configurable background job processing, including concurrency, scheduling, batching, heartbeats, stuck-job detection, and backlog warnings.
  - Added configurable retry policies with maximum runtime, attempt limits, exponential backoff, and delay caps.
  - Added per-job-type retry overrides with fallback to global settings.
- **Validation**
  - Added configuration validation for positive values, compatible heartbeat and stuck-job thresholds, and valid retry backoff ranges.
  - Invalid settings now report the relevant configuration keys and accumulated validation issues.


src/Endatix.Modules.Jobs/Runtime/BackgroundJobRetryPolicy.cs
```diff
@@ -0,0 +1,37 @@
+namespace Endatix.Modules.Jobs.Runtime;
+
+/// <summary>
+/// Decides when a job whose attempt failed retryably becomes eligible again.
+/// </summary>
+internal static class BackgroundJobRetryPolicy
+{
+    // Bounds the doubling so the arithmetic stays finite for any attempt count. At this exponent even a
+    // one-second base already exceeds thirty years, past any cap worth configuring.
+    private const int MaxExponent = 30;
```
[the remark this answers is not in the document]
igdragoev: Fixed in 5452e78e. `MaxExponent` is now 31, so a one-second base can reach any cap the options accept, up to `int.MaxValue` seconds. `NextAttemptAt_OneSecondBaseAndLargestCap_ReachesCap` resolves the policy from `BackgroundJobsOptions` (base 1, cap `int.MaxValue`) and checks attempts 32 and `int.MaxValue` land on the cap. It failed with the old limit and passes now.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobsOptions.cs
```diff
@@ -0,0 +1,166 @@
...
+/// <para>
+/// <see cref="MaxAttempts"/>, <see cref="MaxRuntimeMinutes"/>, <see cref="BackoffBaseSeconds"/> and
+/// <see cref="BackoffCapSeconds"/> apply to every job type unless <see cref="JobTypes"/> overrides them, one
+/// value at a time. The runtime resolves these four values per job type, so reading one of these properties
+/// directly ignores any override in <see cref="JobTypes"/>.
+/// </para>
+/// <para>
+/// Every other value describes the process that runs jobs rather than a kind of job, so it is global only.
+/// </para>
+/// </remarks>
```
PeterMarinov: _⚠️ Potential issue_ | _🟡 medium_ | **XML restates names**

These types are hard to read because almost every member has a paragraph that restates the identifier. Same pattern on the retry policy, validator, and both test classes.

Keep only what names cannot say:

- retry keys must go through `ResolvePolicy` (direct property reads ignore `JobTypes`)
- `JobTypes` keys are case-insensitive; job type strings in the DB are not
- stuck threshold ≥ 3 heartbeats
- first failure is claimed attempt 1 → wait is `base`, not `2 * base`

Drop class-level novels on the tests, the `MaxExponent` essay, and the “does not throw” remarks. Names first; one sentence when there is a real invariant.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobRetryPolicy.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <remarks>
+    /// The exponent is the claimed attempt minus one. The attempt count goes up when a job is claimed, not when
+    /// it fails, so a first attempt that fails is already attempt 1 and has to wait exactly the base delay. An
+    /// attempt below 1 is treated as the first attempt, so it waits the base delay rather than a shorter one and
+    /// does not throw.
+    /// </remarks>
+    public static DateTime NextAttemptAt(int claimedAttempt, DateTime utcNow, BackgroundJobTypePolicy policy)
+    {
+        // Compared before subtracting, so an attempt of int.MinValue cannot wrap round to the largest exponent.
+        var exponent = claimedAttempt <= 1 ? 0 : Math.Min(claimedAttempt - 1, MaxExponent);
```
PeterMarinov: _⚠️ Potential issue_ | _🟠 major_ | **Unbounded ints force a policy for impossible attempts**

`RequireAtLeastOne` allows `MaxAttempts` / `BackoffCapSeconds` up to `int.MaxValue`. That is why this method has `MaxExponent`, double arithmetic, an `int.MinValue` guard, and a test that `claimedAttempt == int.MaxValue` with a 1s base reaches the cap.

Jobs will run with `MaxAttempts` in the single digits (defaults 3 / 8). The policy should be the four-line capped exponential for that domain.

Prefer a small upper bound in the validator (e.g. attempts ≤ 100, backoff in a sane second range). Then `MaxExponent`, the MinValue theory, and `NextAttemptAt_OneSecondBaseAndLargestCap_ReachesCap` can go.

Need unbounded attempts? Say why. Right now this looks like the CodeRabbit chase (`MaxExponent` 30 → 31), not an operator need.

_Uncertain:_ confirm nobody already planned “unlimited retries” via `MaxAttempts = int.MaxValue`.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobsOptions.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// Overrides keyed by job type, for example
+    /// <c>Endatix:BackgroundJobs:JobTypes:SubmissionExport:MaxAttempts</c>. A value left unset falls back to
+    /// the global value of the same name.
+    /// </summary>
+    /// <remarks>
+    /// A key matches its job type regardless of case, even though job types themselves are case-sensitive.
+    /// Configuration keys are case-insensitive everywhere else, so an override written in a different case has
+    /// to apply rather than be silently ignored. An assigned dictionary is copied into one that ignores case,
+    /// whatever comparer it was created with, so assigning a dictionary whose keys differ only in case throws.
+    /// </remarks>
```
PeterMarinov, line 67: _⚠️ Potential issue_ | _🟡 medium_ | **Unknown per-type keys fail closed-silent**

`JobTypes:{Type}:BacklogWarningMinutes` and `JobTypes:{Type}:MaxConcurrency` bind as unknown properties and are ignored. Validation succeeds (`Bind_BacklogWarningMinutes_DefaultsTo15AndIgnoresPerTypeKey`, `Validate_UnknownSections_Succeeds`).

Operators will put concurrency / backlog under the job type — that is the obvious JSON shape. They will not get an error.

Either:
- reject unknown keys under `JobTypes:*` (stricter, better DX), or
- one sentence here: only the four nullable fields override; everything else is ignored on purpose.

Do not add a `Classes` sibling in tests unless that section is a real contract. It currently documents a hole, not a feature.

_Uncertain:_ #1051 says unknown overrides for undeployed **handlers** must pass. That is job types, not unknown **keys** on a type. Those can be split.

tests/Endatix.Modules.Jobs.Tests/Runtime/BackgroundJobsOptionsTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            "backoff cap below base",
+            new() { ["BackoffBaseSeconds"] = "30", ["BackoffCapSeconds"] = "10" },
+            ["BackoffCapSeconds"]
+        },
+        {
+            "job type attempts below 1",
+            new() { ["JobTypes:X:MaxAttempts"] = "0" },
+            ["JobTypes:X:MaxAttempts"]
+        },
+    };
```
PeterMarinov, line 51: _🛠️ Refactor suggestion_ | _🟡 medium_ | **Missing mixed backoff inheritance**

`ValidateJobType` has a real branch: override only cap (or only base) and compare against the inherited global. Messages must name the key the value **came from**.

`InvalidValues` covers global cap &lt; base and `JobTypes:X:MaxAttempts = 0`. It does not cover:
- `JobTypes:X:BackoffCapSeconds` &lt; global base
- `JobTypes:X:BackoffBaseSeconds` &gt; global cap

That is the code most likely to rot. Add those two cases here and assert both full keys in the failure text.

tests/Endatix.Modules.Jobs.Tests/Runtime/BackgroundJobsOptionsTests.cs
```diff
@@ -0,0 +1,246 @@
...
+    [Fact]
+    public void Validate_UnknownSections_Succeeds()
+    {
+        // Arrange — overrides for a job type whose handler is not deployed yet, or a section another component
+        // reads, are not mistakes.
+        var options = Bind(new()
+        {
+            ["JobTypes:UnknownType:MaxAttempts"] = "5",
+            ["Classes:Heavy:MaxConcurrency"] = "2",
+        });
```
PeterMarinov: _⚠️ Potential issue_ | _🟡 medium_ | **`Classes:Heavy` is not a contract**

This case is useful for “unknown job type override is not an error”. The `Classes:Heavy:MaxConcurrency` sibling is not. Workload classes are deferred (#1051 / e862 §5.2). Leaving this key in the suite documents a hole agents will copy.

Keep `JobTypes:UnknownType:MaxAttempts`. Drop `Classes` unless J3 is about to bind that section.

PeterMarinov (changes requested):
**Review** · issues only · host unchanged, but this is the public `Endatix:BackgroundJobs` contract for #1051.

**Merge:** request changes. Slice is the right cut (bind + validate + backoff before the runner). XML that restates names plus unbounded ints will freeze the wrong shape.

**Before merge**
- Trim XML/remarks to the few invariants names cannot say.
- Bound `MaxAttempts` (and likely backoff seconds) *or* keep unbounded ints — not both plus `MaxExponent` / `int.MinValue` theater.
- Cover mixed backoff inheritance in `InvalidValues`.
- Reject unknown keys under `JobTypes:*`, or one sentence that only the four nullables override.
- Pin the options pattern in `oss/ARCHITECTURE.md` (~8 lines). No Docusaurus settings page until something actually registers these options.

**Agent docs (no file in this diff):** `ARCHITECTURE.md` / `AGENTS.md` do not mention Jobs. Do not invent a user settings page. Do pin: section `BackgroundJobsOptions.SectionName`; later `AddOptions` + this validator + `ValidateOnStart`; retry/runtime only via `ResolvePolicy`; `JobTypes` ordinal-ignore-case + nullable per-key fallback; validator messages use the full configuration key. POCO defaults are the spec — do not duplicate them in prose.

**Skipped:** `RetentionDays` (out of scope on #1051), `appsettings` sample while unbound, `DateTime` Kind, `NotThrow()` wrappers.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobsOptions.cs
igdragoev: Done in c83b0e03. Dropped the class summaries on both test classes and on the options, validator and policy types, the ten member paragraphs that restated their own names, the param tags and the "does not throw" remarks. Kept the four invariants you named, plus one sentence each on `RunInProcess`, `BackoffBaseSeconds` and `BacklogWarningMinutes` where the name alone is misleading.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobRetryPolicy.cs
igdragoev: Agreed, and done in c83b0e03 — both halves, because they fit together. `MaxExponent`, the `int.MinValue` theory row and `NextAttemptAt_OneSecondBaseAndLargestCap_ReachesCap` are gone; the method is now the exponent, one `Math.Min` and the return. `Math.Min` already yields the cap when the doubling overflows to infinity, so the only guard left treats an attempt below 1 as attempt 1.

The validator now bounds every value above as well as below, per-type overrides included, with the same ceiling as the global value they replace: `MaxAttempts` 10, `MaxConcurrency` 1000, `SweepBatchSize` 10000, intervals 3600s, thresholds and runtime 1440 minutes, backoff 86400s.

On your uncertainty: no, nobody planned unlimited retries. Dead-lettering at `MaxAttempts` is the design throughout #1051, so 10 costs nothing. The bounds also matter later — a value near `int.MaxValue` breaks `Task.Delay` and the date arithmetic the sweeper and runner will use.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobsOptions.cs
igdragoev: Split as you suggested. Unknown job types keep passing (a handler that is not deployed yet is not a mistake); unknown keys on a type are now documented on `JobTypes` as ignored — only the four nullable values override.

Rejecting them needs the bound configuration section, and nothing binds it until the task that registers the runner and sweeper, so the rejection lands there rather than here. Recorded against that task so it is not lost.

tests/Endatix.Modules.Jobs.Tests/Runtime/BackgroundJobsOptionsTests.cs
igdragoev: Added in c83b0e03: a per-type `BackoffCapSeconds` below the inherited global base, and a per-type `BackoffBaseSeconds` above the inherited global cap. Each asserts both full keys in the failure text.

Worth noting the assertions now check the full key rather than the bare name — matching on `BackoffBaseSeconds` alone also matched inside the per-type key, so those two cases would have proved nothing.

tests/Endatix.Modules.Jobs.Tests/Runtime/BackgroundJobsOptionsTests.cs
igdragoev: Dropped in c83b0e03, from the test and from the criterion behind it. `JobTypes:UnknownType:MaxAttempts` stays as the real rule. Workload classes are commercial and deferred, so nothing binds a `Classes` section.

igdragoev:
Thanks — all five addressed in c83b0e03, and the options contract is pinned in `ARCHITECTURE.md` (8 lines under module registration): section named by the options type, `ValidateOnStart` with the validator when the module registers them, full configuration keys in validation messages, the case-insensitive per-key nullable fallback, and that overridable values are read only through `ResolvePolicy`. No defaults restated in prose and no documentation-site page.

Deferred on purpose, each recorded so it is not lost: rejecting unknown keys under `JobTypes:*` belongs to the task that binds the section; `RetentionDays`, an `appsettings` sample and the workload-class section stay out until something registers or reads them.

Net effect on the diff: 83 added, 139 removed.

PeterMarinov (approved):
Approved

PR #1058, opened 2026-09-16
feat(#1059): add the background job state repository

### LINKED ISSUE #1059: Add a fenced state repository for background jobs
### Goal

The job runner and sweeper need to change a job's state safely when several instances, and several attempts of the same job, can reach the same row. Add an internal state repository to the Jobs module whose writes are fenced, so a stale or superseded runner cannot overwrite newer state.

### Implementation Notes

- Internal to `Endatix.Modules.Jobs`, with no callers yet. The runner and sweeper that use it are tracked in #1051.
- Every write is a single set-based update, gated on the status it expects and on the claimed attempt, and reports whether exactly one row changed.
- Claiming is gated on status, due time and a registered job type, and consumes the attempt at claim. Two concurrent claims of one job produce exactly one winner.
- A failed attempt retries or dead-letters in the same update, depending on whether attempts remain. Reaping re-checks the stale heartbeat in the statement that writes.
- The discovery reads, for eligible and for stale jobs, are ordered, capped, read-only and span tenants.
- Error messages are truncated to the column length, and the exception-text guard from #1049 covers the new failure writers.

igdragoev:
Closes #1059. Part of #1051.

The job runner and sweeper need to change a job's state safely when several instances, and several attempts of the same job, can reach the same row. This adds that repository, internal to the Jobs module. Nothing calls it yet.

- Claiming is a read followed by a compare-and-swap. The update is gated on status, due time and a registered job type, and on the attempt count still being the one just read. A successful claim returns the job with the attempt it took, and that attempt fences every later write. A runner never learns its attempt from a separate read, so a job that was reaped and claimed again by another runner cannot be mistaken for its own.
- Every later write is a single set-based update, fenced on the status it expects and on the claimed attempt. A runner whose attempt was reaped or superseded changes nothing and learns so from the returned `false`.
- Two concurrent claims of one job produce exactly one winner.
- A failed attempt either retries or dead-letters in the same update, depending on whether attempts remain. Reaping re-checks staleness in the statement that writes, and a running job with no heartbeat counts as stale, since nothing shows its runner is alive.
- The discovery reads, for eligible and for stale jobs, are ordered, capped, read-only and span tenants. Expiry does not affect eligibility: it is retention, and the retention collector removes terminal rows only.
- Failure messages must be non-empty, as on the entity, and are cut to the column length. The exception-text guard also covers the two new failure writers.

Cancelling a job and reporting a running job's progress through conditional writes is #1061.

The Jobs module grants its internals to NSubstitute's proxy assembly by name alone, because proxies of its own internal interface are unsigned. Integration tests now also reset the `jobs` schema between tests.


src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,250 @@
...
+                setters => setters
+                    .SetProperty(job => job.Status, JobStatus.Processing)
+                    .SetProperty(job => job.AttemptCount, job => job.AttemptCount + 1)
+                    // Kept rather than overwritten: this records when the job began, not when the
+                    // current attempt did, so a retry must not reset it.
+                    .SetProperty(job => job.StartedAt, job => job.StartedAt ?? utcNow)
+                    .SetProperty(job => job.HeartbeatAt, (DateTime?)utcNow),
+                cancellationToken);
+
+        return affected == 1;
```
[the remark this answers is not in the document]
igdragoev: Fixed in 6d935498, provider-agnostic rather than `UPDATE ... RETURNING`: EF's `ExecuteUpdateAsync` can't return columns, and this module is meant to gain a SQL Server provider.

`TryClaimAsync` now reads the row, then runs the conditional update with one more condition: `AttemptCount` still equals the value just read. On success it returns the `ClaimedJob` with `AttemptCount + 1`, otherwise `null`. The fence comes from that compare-and-swap, and `GetClaimedAsync` is removed, so nothing can learn the attempt by job id alone. If another claim changes the count between the read and the swap, the swap matches nothing and the job is rediscovered on the next sweep.

`TryClaimAsync_AttemptCountChangedSinceRead_ClaimsNothing` covers a swap with a stale count, and fails without the new condition.

src/Endatix.Core/Endatix.Core.csproj
```diff
@@ -26,6 +26,6 @@
 	</ItemGroup>
 	<ItemGroup>
 		<InternalsVisibleTo Include="Endatix.Core.Tests" />
-		<InternalsVisibleTo Include="DynamicProxyGenAssembly2" PublicKey="$(NSubstitutePublicKey)" />
+		<InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
```
PeterMarinov: _🔒 Security / blast radius_ | _🟠 Major_ | _⚡ Quick win_

**Don't widen Core internals for a Jobs substitute.**

This assembly was key-qualified (`PublicKey=$(NSubstitutePublicKey)`). Jobs only needs unsigned `DynamicProxyGenAssembly2` for `IBackgroundJobStateRepository` — that grant belongs on `Endatix.Modules.Jobs.csproj` alone.

Same change is on `Endatix.Infrastructure.csproj` and `Directory.Build.props` (property removed). Revert all three unless a Core/Infra substitute actually fails with the keyed grant (paste the test).

<details>
<summary>🤖 Prompt for AI Agents</summary>

Restore `PublicKey="$(NSubstitutePublicKey)"` on Core and Infrastructure `InternalsVisibleTo` for `DynamicProxyGenAssembly2`. Restore `NSubstitutePublicKey` in `Directory.Build.props`. Keep the unsigned grant only on the Jobs module csproj.
</details>

src/Endatix.Infrastructure/Endatix.Infrastructure.csproj
```diff
@@ -53,6 +53,6 @@
 	<ItemGroup>
 		<InternalsVisibleTo Include="Endatix.Infrastructure.Tests" />
 		<InternalsVisibleTo Include="Endatix.IntegrationTests" />
-		<InternalsVisibleTo Include="DynamicProxyGenAssembly2" PublicKey="$(NSubstitutePublicKey)" />
+		<InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
```
PeterMarinov: _🔒 Security / blast radius_ | _🟠 Major_ | _⚡ Quick win_

**Same as Core:** dropping the public key here is unrelated to the Jobs state repository. Revert unless Infrastructure tests require an unsigned Castle proxy for an *Infrastructure* internal type (they don't in this PR).

<details>
<summary>🤖 Prompt for AI Agents</summary>

Restore `PublicKey="$(NSubstitutePublicKey)"` on this `InternalsVisibleTo` entry.
</details>

src/Endatix.Modules.Jobs/Runtime/JobDispatchItem.cs
```diff
@@ -0,0 +1,12 @@
...
+/// <summary>
+/// One piece of work offered to the runner: which job, and of what type.
+/// </summary>
+/// <remarks>
+/// The job type travels with the id so that whatever routes the item — a single pool today, a
+/// per-workload-class strategy later — can decide where it goes without reading the row back.
+/// </remarks>
+/// <param name="JobId">Identifier of the job row.</param>
+/// <param name="JobType">The job's router key, compared ordinal and case-sensitive.</param>
+public readonly record struct JobDispatchItem(long JobId, string JobType);
```
PeterMarinov: _🧹 Code quality_ | _🟡 Minor_ | _⚡ Quick win_

**`JobDispatchItem` should be `internal` (same as `ClaimedJob` / `IBackgroundJobStateRepository`).**

It's only the FindEligible DTO. `public` ships on the NuGet package before a runner exists. Rules: internal by default. Param names are enough — drop the remarks.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Change `public readonly record struct JobDispatchItem` to `internal`. Trim the XML remarks.
</details>

src/Endatix.Modules.Jobs/Endatix.Modules.Jobs.csproj
```diff
@@ -37,5 +37,11 @@
   <ItemGroup>
     <InternalsVisibleTo Include="Endatix.Modules.Jobs.Tests" />
     <InternalsVisibleTo Include="Endatix.IntegrationTests" />
+    <!-- NSubstitute proxies are emitted into this assembly, so without a grant a substitute for an
+         internal type fails at run time rather than at compile time. The grant names no public key.
+         The proxy assembly is signed only when the proxied type comes from a strong-named assembly,
+         as ILogger<T> does, so a key-qualified grant never matches the proxy of an internal interface
+         declared here. This assembly is not strong-named, so a grant by name alone matches both. -->
+    <InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
```
PeterMarinov, line 41: _📝 Documentation_ | _🟡 Minor_ | _⚡ Quick win_

**This unsigned grant is the right Jobs-only fix. Don't copy it to Core/Infrastructure.**

Also: csproj `Description` (unchanged in this hunk) still sells a sweeper and bounded runner; README still says enqueue-only / no runner. One-line truth: state repository exists, unused, not hosted. Don't add an AGENTS.md chapter.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Keep this Jobs `InternalsVisibleTo` unsigned. Align package Description + README with current shipping surface. Do not document the runner in AGENTS.md.
</details>

src/Endatix.Modules.Jobs/Runtime/IBackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,153 @@
...
+/// </para>
+/// <para>
+/// The current time is a parameter rather than something read here, so one tick decides a single
+/// instant for all the rows it touches and tests can drive the clock.
+/// </para>
+/// <para>
+/// Reads run with the tenant query filter left on. Outside a request the ambient tenant is 0, which
+/// makes that filter permissive — that is how the sweeper sees every tenant's jobs without holding a
+/// tenant scope of its own.
+/// </para>
```
PeterMarinov: _🧹 Code quality_ | _🟡 Minor_ | _🏗️ Heavy lift_

**Cut the novels. One fence rule, once.**

Keep: ExecuteUpdate (no SaveChanges / ModifiedAt), claim CAS on attempt, later writes fenced on Processing + attempt, utcNow injected, tenant 0 = filter off.

Delete restated remarks on every method. Same diet on `BackgroundJobStateRepository` `//` comments and most integration-test remarks — keep the one-liner that states the invariant.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Compress this interface to signatures plus a short type-level remark (5–8 lines). Strip duplicate `<remarks>` on individual methods.
</details>

src/Endatix.Modules.Jobs/Runtime/IBackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// <summary>
+    /// Finds jobs that are due to run and whose type this instance can handle, earliest first. Changes
+    /// nothing.
+    /// </summary>
+    Task<IReadOnlyList<JobDispatchItem>> FindEligibleAsync(
+        IReadOnlyCollection<string> registeredJobTypes,
+        DateTime utcNow,
+        int limit,
+        CancellationToken cancellationToken = default);
+}
```
PeterMarinov, line 68: _⚠️ Potential issue_ | _🟡 Minor_ | _❓ Needs confirmation_

**No fenced progress or cancel.**

Domain has `ReportProgress` and `Cancel`. This type has neither. If this is the only safe writer, cancel-from-Processing and progress belong here (or an explicit follow-up issue). Otherwise the next API cancel will `SaveChanges` and fight `HeartbeatAt`.

Confirm against #1051 / runner: this PR vs next. Don't leave it implicit.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Either add fenced `TryCancel` / `TryReportProgress` with PG tests, or file/link an issue and one README sentence that entity Cancel/progress is still TODO.
</details>

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,278 @@
...
+        if (alsoWhere is not null)
+        {
+            fenced = fenced.Where(alsoWhere);
+        }
+
+        return await fenced.ExecuteUpdateAsync(setters, cancellationToken) == 1;
+    }
+
+    private static string Truncate(string errorMessage) =>
+        errorMessage.Length <= ErrorMessageMaxLength ? errorMessage : errorMessage[..ErrorMessageMaxLength];
```
PeterMarinov: _🐛 Bug_ | _🟠 Major_ | _⚡ Quick win_

**`Truncate` NRE on null; empty/whitespace allowed here, rejected on the entity.**

`errorMessage.Length` throws if null. `BackgroundJob.Fail` / `DeadLetter` use `Guard.Against.NullOrWhiteSpace`. The SQL path can store `""` / `"   "` on the job-status API.

Match the domain guard (or coerce), then slice to 2048. Add a PG test on `RecordFailedAttemptAsync` / `TryReapAsync` truncation — today only `TryFailAsync` is covered.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Null-safe truncate; reject or coerce null/whitespace like the entity; integration test the shared `FailedAttemptSetters` path.
</details>

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,278 @@
...
+            job => job.HeartbeatAt < staleCutoff);
+
+    /// <inheritdoc />
+    public async Task<IReadOnlyList<StaleJob>> FindStaleAsync(
+        DateTime staleCutoff,
+        int limit,
+        CancellationToken cancellationToken = default) =>
+        await dbContext.BackgroundJobs
+            .AsNoTracking()
+            .Where(job => job.Status == JobStatus.Processing && job.HeartbeatAt < staleCutoff)
```
PeterMarinov: _⚠️ Potential issue_ | _🔵 Minor_ | _❓ Needs confirmation_

**Stale scan: `HeartbeatAt < cutoff` ignores NULL.**

SQL `NULL < date` is unknown → row excluded. Claim sets heartbeat; a Processing row with null heartbeat (bad seed / future writer) is never reaped.

Should Processing + `HeartbeatAt IS NULL` count as dead? If yes: `job.HeartbeatAt == null || job.HeartbeatAt < staleCutoff` + a test. If no: document “claim always sets heartbeat; null in-flight is invalid”.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Decide the NULL rule, implement or document, add an integration case if NULL is stale.
</details>

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,278 @@
...
+        if (registeredJobTypes.Count == 0)
+        {
+            return [];
+        }
+
+        return await dbContext.BackgroundJobs
+            .AsNoTracking()
+            .Where(job => (job.Status == JobStatus.Pending || job.Status == JobStatus.Retrying)
+                && job.NextAttemptAt <= utcNow
+                && registeredJobTypes.Contains(job.JobType))
```
PeterMarinov, line 164: _⚠️ Potential issue_ | _🔵 Minor_ | _❓ Needs confirmation_

**Eligible query does not filter `ExpiresAt`.**

Column + index exist. Dispatch still returns expired rows.

If expiry is retention-only (410 on download), say so next to FindEligible. If expired must not run, add `ExpiresAt == null || ExpiresAt > utcNow` and a test.

<details>
<summary>🤖 Prompt for AI Agents</summary>

Either add the expiry predicate + PG test, or a one-line comment that ExpiresAt is retention-only and must not affect dispatch.
</details>

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
```diff
@@ -0,0 +1,278 @@
...
+    /// <summary>
+    /// Claims the job only while its attempt count is still <paramref name="expectedAttemptCount"/>.
+    /// </summary>
+    /// <remarks>
+    /// Only a claim changes the attempt count, so a claim that landed after the caller read the row makes
+    /// this update match nothing, and the attempt the caller is handed can be built from its own read.
+    /// </remarks>
+    /// <returns><c>true</c> when this update took the job, which then holds attempt
+    /// <paramref name="expectedAttemptCount"/> + 1.</returns>
+    internal async Task<bool> TryClaimAtAttemptAsync(
```
PeterMarinov: _🧹 Code quality_ | _🔵 Minor_ | _⚡ Quick win_

**Prefer one claim entry.**

`internal TryClaimAtAttemptAsync` is a second claim API so tests can skip the read. Callers could bypass reconstruct-from-read.

Test CAS via two connections + `TryClaimAsync`, or keep the helper `private`.

Also worth adding if you keep the rest: `TryReapAsync` when `claimedAttempt >= maxAttempts` → DeadLettered (retry path only today).

<details>
<summary>🤖 Prompt for AI Agents</summary>

Hide or delete `TryClaimAtAttemptAsync` from the type's test surface; assert attempt-changed CAS through `TryClaimAsync` if possible.
</details>

PeterMarinov (changes requested):
_🎯 Functional Correctness_ | _🟠 Major_

CAS claim + fenced `ExecuteUpdate` + PG race tests are the right slice. **Not merge-ready.**

1. Revert Core/Infrastructure `InternalsVisibleTo` public-key removal (and `NSubstitutePublicKey` in `Directory.Build.props`) — Jobs-only unsigned grant.
2. `JobDispatchItem` `internal`.
3. Package `Description` still sells a runner/sweeper; README still says enqueue-only — align (README not in this diff).
4. `Truncate(null)` NRE; empty/whitespace vs domain `Fail`/`DeadLetter` guards.
5. Cut XML remarks (~70%).
6. Confirm cancel/progress, `ExpiresAt`, null `HeartbeatAt` — this PR or a linked issue.

Do not add in-memory tests for `ExecuteUpdate`.

src/Endatix.Core/Endatix.Core.csproj
igdragoev: Reverted in d04b4da1 — the keyed grant and `NSubstitutePublicKey` are restored, byte-identical to main. You're right: Core and Infrastructure tests substitute `ILogger<internal type>`, whose proxy is signed, and the keyed grant covers that. Only Jobs proxies its own internal interface, so the unsigned grant stays on the Jobs csproj alone.

src/Endatix.Infrastructure/Endatix.Infrastructure.csproj
igdragoev: Same as Core — reverted in d04b4da1.

src/Endatix.Modules.Jobs/Endatix.Modules.Jobs.csproj
igdragoev: Kept Jobs-only (d04b4da1). The package `Description` and the README now say the same thing (e8c8a7c2): a state repository exists, nothing uses it yet, and no runner or sweeper is hosted. Nothing added to AGENTS.md.

src/Endatix.Modules.Jobs/Runtime/JobDispatchItem.cs
igdragoev: Internal now, remarks dropped (91debce6). It becomes public together with `IJobDispatchStrategy`, the seam that needs it.

src/Endatix.Modules.Jobs/Runtime/IBackgroundJobStateRepository.cs
igdragoev: Cut in 91debce6 to signatures plus one type-level remark carrying the five points you listed. The per-method remarks, the implementation comments and the test remarks got the same treatment.

src/Endatix.Modules.Jobs/Runtime/IBackgroundJobStateRepository.cs
igdragoev: Out of this PR: filed #1061 for conditional cancel and progress writes, which belong with the cancel endpoint and the progress reporter, and the README points to it (e8c8a7c2). Agreed that an entity `Cancel` + `SaveChanges` would be unsafe against a runner that completes the job in between.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
igdragoev: Fixed in 91debce6: the failure writers now guard `NullOrWhiteSpace` like `Fail`/`DeadLetter`, then cut to 2048. Added PostgreSQL tests for truncation through `RecordFailedAttemptAsync` and `TryReapAsync`, and for null, empty and whitespace messages being rejected without a write.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
igdragoev: Counted as dead (91debce6): a `Processing` row with no heartbeat has nothing proving its runner alive. The stale scan and the reap both include it, those rows come first, and `FindStaleAsync_NullHeartbeat_ReturnsFirstAndReaps` covers it.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
igdragoev: Retention only, so it's not filtered here — documented next to `FindEligibleAsync` (91debce6). Filtering it would leave expired `Pending` rows that the terminal-only collector never removes.

src/Endatix.Modules.Jobs/Runtime/BackgroundJobStateRepository.cs
igdragoev: The swap helper is private (91debce6). The stale-count case now goes through `TryClaimAsync`: a second connection holds the row lock, moves the attempt count on once the claim's update is waiting on that lock, and the claim must return `null` (`TryClaimAsync_AttemptCountChangedSinceRead_ReturnsNull`). Also added `TryReapAsync_AttemptsExhausted_DeadLetters`.

igdragoev:
Thanks — all addressed except fenced cancel/progress, which is now #1061: it belongs with the cancel endpoint and the progress reporter, and the README points to it.

- Grants: Core and Infrastructure reverted to the keyed grant (d04b4da1); the unsigned grant stays Jobs-only.
- Repository (91debce6): `JobDispatchItem` internal; remarks cut; the swap helper is private and the stale-count case goes through `TryClaimAsync` with a row lock on a second connection; failure messages guarded like the entity and truncation covered on every failure writer; reap to dead-letter covered.
- Package `Description` and README aligned with what ships (e8c8a7c2).

On the two confirm questions: a `Processing` row with a null heartbeat is now treated as stale and reaped, and `ExpiresAt` stays retention-only, documented at `FindEligibleAsync`, because filtering it out of dispatch would strand `Pending` rows the terminal-only collector never removes. Tests stay PostgreSQL-only.

PeterMarinov (approved):
Approved

PR #1067, opened 2026-09-18
feat(#1066): offer committed jobs to the local runner after enqueue

### LINKED ISSUE #1066: Offer newly enqueued jobs to the local runner
### Goal

When a job is enqueued on an instance that runs jobs, offer it to that instance's runner straight away, so it starts without waiting for the next sweep.

### Implementation Notes

- The offer happens only after the job row is committed, in the order the jobs were requested. A failed save offers nothing.
- It is a latency optimisation: a rejected or failed offer leaves the job for the next sweep, and the caller's result never changes.
- Enqueued jobs and rejected offers are recorded through `IJobMetrics`.
- The dispatch strategy and metrics are optional, so a host that registers neither enqueues exactly as before. Registration arrives with the runner (#1051).

igdragoev:
Closes #1066. Part of #1051.

Enqueueing now tells the local runner about new jobs straight away, instead of leaving them for the next sweep. It is a latency optimisation: the job row is already committed, so a missed offer only means the sweeper finds the job a little later.

- After `SaveChangesAsync` succeeds, the queue offers each job to the dispatch strategy in the order it was requested, then records `Enqueued` for each job and `OfferRejected` for each rejected offer. Offers come first so that a slow metrics sink cannot delay dispatch, and one failed offer does not stop the rest.
- If the save fails, nothing is offered or recorded, and the exception reaches the caller.
- The strategy, the metrics sink and the logger are optional, so enqueueing works as before where none is registered, which is every host today.
- An exception from the strategy or the metrics is logged and swallowed. It never reaches the caller or changes the ids the caller gets.
- The module README and package description now say that enqueueing calls the extension points when a host registers them.

## Summary by CodeRabbit

- **New Features**
  - Newly enqueued background jobs can be offered for immediate dispatch when a host registers a dispatch strategy.
  - Job lifecycle metrics now record enqueue activity and dispatch outcomes.
  - Enqueue operations continue returning job IDs when dispatch is rejected or registered integrations encounter errors.

- **Bug Fixes**
  - Jobs are offered only after successful database commits, preventing dispatch of jobs that were not saved.

- **Documentation**
  - Updated package and README descriptions to clarify supported extension points and current usage.


src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
```diff
@@ -80,4 +92,62 @@ private BackgroundJob CreateJob(BackgroundJobRequest request)
...
             traceId: Activity.Current?.Id);
     }
+
+    // The rows are already committed, so nothing from here on may throw into the caller or change the ids it gets.
+    private void SignalCommitted(IReadOnlyList<BackgroundJob> jobs)
+    {
+        foreach (var job in jobs)
+        {
+            Record(JobLifecycleEvent.Enqueued, job.JobType);
+        }
```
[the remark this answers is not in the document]
igdragoev: Agreed — fixed in 45092ca1. The batch is now offered first, in request order; `Enqueued` and `OfferRejected` are recorded afterwards, each still guarded so an exception is logged and swallowed. A new test logs the call order and fails if any metric call precedes an offer, and another checks that a throwing first offer doesn't stop the rest.

src/Endatix.Modules.Jobs/README.md
```diff
@@ -29,8 +29,9 @@ second system can disagree with it about what a job is doing.
 
 > [!IMPORTANT]
 > This package currently provides **persistence and enqueueing**. It also contains an internal
-> job state repository and the public `IJobDispatchStrategy` and `IJobMetrics` extension points,
-> which nothing uses yet. No runner or sweeper is hosted, so enqueued jobs stay `Pending`. The
+> job state repository, which nothing uses yet, and the public `IJobDispatchStrategy` and
+> `IJobMetrics` extension points, which enqueueing calls only when a host registers them. No
+> runner or sweeper is hosted, so enqueued jobs stay `Pending`. The
```
PeterMarinov, line 34: _📚 Docs / agent drift_ | _🟡 Medium_

**This box now says enqueue calls the seams; Runtime below still does not.**

Here (and the package description) enqueueing calls `IJobDispatchStrategy` / `IJobMetrics` when a host registers them. Runtime (~L132) still says only the runner/sweeper will pass work through those seams. Enqueueing still reads as insert-only.

Agents will pick one paragraph. Pin one pattern, short:

1. Commit the row.
2. If a strategy is registered, `TryOffer` each id (latency only; reject/throw → sweeper).
3. If metrics are registered, record `Enqueued` then `OfferRejected`.
4. Module still registers neither; without a runner, rows stay `Pending` even if offered into a channel.

Drop “runner and sweeper will…” as the only call sites. Don’t add a second essay.

_Needs verify:_ Docusaurus has no Jobs page — confirm we are not supposed to add one for this slice.

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
```diff
@@ -12,21 +15,26 @@ namespace Endatix.Modules.Jobs.Features;
...
+/// reaches the caller.
 /// </para>
 /// </remarks>
 internal sealed class BackgroundJobQueue(
     IJobsDbContext dbContext,
-    IDateTimeProvider dateTimeProvider) : IBackgroundJobQueue
+    IDateTimeProvider dateTimeProvider,
+    IJobDispatchStrategy? dispatchStrategy = null,
+    IJobMetrics? metrics = null,
+    ILogger<BackgroundJobQueue>? logger = null) : IBackgroundJobQueue
```
PeterMarinov, line 34: _🧪 Tests_ | _🟡 Medium_

**Optional `IJobDispatchStrategy` / `IJobMetrics` never resolved through DI.**

Tests construct the queue by hand. Production is `AddScoped<IBackgroundJobQueue, BackgroundJobQueue>()` in `JobsModule`. Optional ctor params work on modern MS.DI only if unregistered services are skipped.

If that fails, turning `JobsModule` on throws on first enqueue — not “works as before.”

Add a `ServiceProvider` test: those registrations + `GetRequiredService<IBackgroundJobQueue>()` with no strategy/metrics registered. Do not add Testcontainers for this.

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
```diff
@@ -80,4 +92,76 @@ private BackgroundJob CreateJob(BackgroundJobRequest request)
...
+                    rejected.Add(job);
+                }
+            }
+            catch (Exception exception)
+            {
+                _logger.LogDebug(
+                    exception,
+                    "Offering background job {JobId} of type {JobType} failed; the next sweep will find it",
+                    job.Id,
+                    job.JobType);
```
PeterMarinov, line 140: _🩺 Observability_ | _🟡 Medium_

**Failed offers/metrics are swallowed at Debug.**

A throwing `TryOffer` is not `OfferRejected` and logs Debug. Default host level is Information, so a broken host strategy is silent: caller still gets ids, job waits for a sweeper that does not exist yet. Same pattern in `Record`.

Either log Warning (with exception) for the throw path, or record a distinct metric. Debug-only is only OK if you explicitly want this invisible until a sweeper exists — call that out.

_Needs verify:_ is Debug intentional so a full in-process channel is not noisy?

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
```diff
@@ -80,4 +92,76 @@ private BackgroundJob CreateJob(BackgroundJobRequest request)
...
+        }
+
+        foreach (var job in rejected)
+        {
+            Record(JobLifecycleEvent.OfferRejected, job.JobType);
+        }
+    }
+
+    // Returns the jobs the strategy refused. An offer that throws is logged rather than counted as refused, and does
+    // not stop the offers after it.
```
PeterMarinov, line 114: _📊 Metrics_ | _🔵 Low_ | _uncertainty_

**Throw vs `false` is a metrics hole.**

`false` → `OfferRejected`. Throw → Debug log, no rejection event. Operators comparing reject rate vs delayed jobs will undercount.

If that split is deliberate (refuse = backpressure, throw = bug), one README clause is enough. If not, count throws as rejected (or a separate event).

_Needs verify:_ issue #1066 / strategy contract — confirm throw is “never” for a correct `TryOffer`.

PeterMarinov (approved):
Post-commit offer is the right latency seam for #1051: save first, then `TryOffer` in request order, then metrics, never fail the caller. Tests cover order, partial throw, reject, save-fail, and no-strategy.

Approving. Follow-ups below are docs/observability, not blockers for this slice.

- README Runtime still says only runner/sweeper use the seams; enqueue already calls them. Enqueueing still reads insert-only.
- Swallowed strategy/metrics exceptions at Debug are invisible at default Information.
- Nothing proves optional ctor params resolve via `AddScoped<IBackgroundJobQueue, BackgroundJobQueue>()`.

No extra integration tests or Docusaurus pages. Production is still a no-op because the module registers neither seam. Value is the pinned contract for the stacked runner work.

src/Endatix.Modules.Jobs/README.md
igdragoev: Addressed in #1071: the README's Runtime section now pins one sequence — commit the rows, offer each job when a strategy is registered, then record `Enqueued` and any `OfferRejected` when metrics are registered — and closes by saying the module registers neither seam yet, so rows stay `Pending` without a runner. Confirmed on the other question: no Docusaurus page for this slice.

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
igdragoev: Added in #1071: `JobsModuleQueueResolutionTests` builds a provider from the module's own registrations and resolves `IBackgroundJobQueue` with no strategy or metrics registered, plus a case that registers both and shows DI passes them through. Removing the optional default makes the first case fail with "Unable to resolve service for type 'IJobDispatchStrategy'".

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
igdragoev: Changed in #1071: a throw from either seam now logs at Warning with the exception. You're right that Debug hid a host bug at the default level. A refusal stays silent, since that is back-pressure rather than a fault.

src/Endatix.Modules.Jobs/Features/BackgroundJobQueue.cs
igdragoev: Deliberate, and now written down in #1071: `false` is back-pressure and is counted as `OfferRejected`; a throw is a bug in the strategy, is logged at Warning and is not counted as a refusal. `TryOffer`'s contract now says a correct strategy never throws — it returns `false` when it has no room — and the shipped strategy cannot throw, because its channel returns `false` when full.

